PR #31, opened 2023-09-19
Lots of goodies for better code

andresuribe87:
This is also coming included now in CI


build.gradle.kts
```diff
@@ -1,3 +1,38 @@
+import io.gitlab.arturbosch.detekt.Detekt
+
 plugins {
-  id("org.jetbrains.kotlin.jvm") version "1.9.0" apply false
+  id("org.jetbrains.kotlin.jvm") version "1.9.0"
```
mistermoe, line 4: was `apply false` removed because we're using this plugin in this gradle file now?

andresuribe87: Correct. It's used for the 
```
  kotlin {
    explicitApi()
  }
```
section. It's applied to all subprojects.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,30 @@
+name: SDK Kotlin CI
+
+on:
```
mistermoe, line 3: can we add  [`workflow_dispatch`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch) so that we can trigger this manually through the UI when needed?

credentials/src/main/kotlin/Validation.kt
```diff
@@ -1,62 +1,57 @@
 package web5.credentials
 
-import java.util.Date
+import java.util.*
```
mistermoe, line 3: what's the thinking behind importing the kitchen sink vs explicit imports?

mistermoe (commented):
left a few comments, otherwise looks great. excited to have stuff in place that lets me know when i'm screwing stuff up, which is quite often

credentials/src/main/kotlin/Validation.kt
andresuribe87: Ah, it's the editorconfig. Fixed this now.

.github/workflows/ci.yml
andresuribe87: Added.

mistermoe (approved):

nitro-neal (approved):
lgtm

PR #33, opened 2023-09-20
adding types to model directory

nitro-neal:
# Overview
https://github.com/TBD54566975/web5-sdk-kotlin/issues/32

Adding types to model directory

# Description
Moving model files to its own class and new directory


nitro-neal:
Even though it is recommended to not use data classes for api
https://kotlinlang.org/docs/jvm-api-guidelines-backward-compatibility.html#don-t-add-arguments-to-existing-api-functions

Something as simple as this:
```
public data class User(
  val id: Int,
  val name: String,
  val email: String?
)
```

Would turn into this
```
class User(
    val id: Int,
    val name: String,
    val email: String?
) {
    // Manually implement equals()
    override fun equals(other: Any?): Boolean {
        if (this === other) return true
        if (javaClass != other?.javaClass) return false

        other as User

        if (id != other.id) return false
        if (name != other.name) return false
        if (email != other.email) return false

        return true
    }

    // Manually implement hashCode()
    override fun hashCode(): Int {
        var result = id
        result = 31 * result + name.hashCode()
        result = 31 * result + (email?.hashCode() ?: 0)
        return result
    }

    // Manually implement toString()
    override fun toString(): String {
        return "User(id=$id, name='$name', email=$email)"
    }
}
```

'Almost any change makes the API not backward-compatible.' I don't think we will really care about this. I have seen a lot of kotlin examples use data classes. I think that even if we dont use data classes we will still make breaking changes so it gains us nothing but more of a headache and more verbose coding.

Thoughts?

andresuribe87 (changes requested):
Why have a separate file per class? The [suggestion](https://kotlinlang.org/docs/coding-conventions.html#source-file-organization) is see explicitly states that 

> Placing multiple declarations (classes, top-level functions or properties) in the same Kotlin source file is encouraged as long as these declarations are closely related to each other semantically, and the file size remains reasonable (not exceeding a few hundred lines).

Can we have only a couple of files?

Re: avoiding dataclasses, I would stick to the guidance. You can use a simple class. No need to override all those methods.

For example
```kotlin
public class SignOptions(
  public val kid: String,
  public val issuerDid: String,
  public val subjectDid: String,
  public val signerPrivateKey: JWK,
)
```

nitro-neal:
A. Well I think how it was before is fine if we want to group things together at a high level  (all ssi stuff lives in SSITypes.kt)
* option 1 - one big file with all ssi types
* option 2 - do what is in PR and everything gets its own file
* option 3 - have 3 files. 

VerifiableCredentialModel.kt
  VerifiableCredentialType
  CredentialStatus
  CredentialSubject
  ...

VerifiablePresentationModel.kt
  VerifiablePresentationType
  ...
  

PresentationExchange.kt
  InputDescriptor
  PresentationDefinitonV2
  ...

I could see it either way actually,. I'm always a fan for everything in the same file with a large grouping, seperated logically by comments or somthing (option1 SSITypes.kt)

But I could definitly see option 3 to have value too.

What do you think?

B. I'll switch it to class and I'll make sure there are no gotchas for that vs data class

mistermoe:
strangely i'm in favor of option 1 🙈

andresuribe87:
1 or 2 for me. 3 is def overkill.

andresuribe87 (approved):

PR #46, opened 2023-09-28
updating pex with pres submission

nitro-neal:
# Overview
Adding more tests for pex and adding more robust pex handling

# Description
Adding      
`generatePresentationSubmissionFrom(createVpOptions.presentationDefinition, createVpOptions.verifiableCredentialJwts): PresentationSubmission` which does field validation for determining if the VCs provided are valid for the presentation definition.

```

/**
   * Validates a list of Verifiable Credentials (VcJwts) against a presentation definition and generates a presentation submission
   *
   * This method checks the presentation definition and the list of Verifiable Credentials to ensure
   * that the required fields in the definition are satisfied by the provided credentials.
   *
   * Currently, if the presentation definition contains submission requirements, or if any required field in an input
   * descriptor is not satisfied by the provided credentials, an exception is thrown.
   *
   * A compliant presentation submission is returned
   *
   * ### Throws
   * - [NotImplementedError] if the Presentation Definition contains Submission Requirements or if a Field Filter is implemented.
   * - [Exception] if any required field is not satisfied in the given InputDescriptor.
   *
   * @param presentationDefinition The [PresentationDefinitionV2] object representing the presentation's definition.
   * @param vcJwts The list of [VcJwt] representing the Verifiable Credentials to be validated against the presentation definition.
   * @throws Exception If a required field is not satisfied in a given InputDescriptor.
   * @return the generated presentation submission.
   */
```

### DIFF
--- a/credentials/src/main/kotlin/SSI.kt
+++ b/credentials/src/main/kotlin/SSI.kt
@@ -150,19 +150,6 @@ public data class CreateVpOptions(
   val holder: String, // TODO: Remove this
 )
 
-/**
- * Represents the decoded parts of a Verifiable Credential JWT.
- *
- * @property header The header of the JWT, containing metadata about the token.
- * @property payload The payload of the JWT, containing the claims and the issuer of the token.
- * @property signature The signature of the JWT, used for verifying the integrity of the token.
- */
-public data class DecodedJwt(
-  val header: Any,
-  val payload: Any,
-  val signature: String,
-)
-
 public typealias VcJwt = String
 public typealias VpJwt = String
 
@@ -241,22 +228,6 @@ public object VerifiableCredential {
     return JwtVerifiableCredential.fromCompactSerialization(vcJWT)
       .verify_Ed25519_EdDSA(publicKeyJWK.toOctetKeyPair())
   }
-
-  /**
-   * Decodes a verifiable credential JWT into its header, payload, and signature components.
-   *
-   * @param vcJWT The JWT representation of the verifiable credential to be decoded.
-   * @return A [DecodedJwt] object containing the decoded components.
-   */
-  public fun decode(vcJWT: VcJwt): DecodedJwt {
-    val (encodedHeader, encodedPayload, encodedSignature) = vcJWT.split('.')
-
-    return DecodedJwt(
-      header = String(Base64.getDecoder().decode(encodedHeader)),
-      payload = String(Base64.getDecoder().decode(encodedPayload)),
-      signature = encodedSignature
-    )
-  }
 }
 
 /**
@@ -278,9 +249,10 @@ public object VerifiablePresentation {
       createVpOptions.verifiableCredentialJwts
     )
 
-    val properties: MutableMap<String, Any> = HashMap()
-    properties["verifiableCredential"] = createVpOptions.verifiableCredentialJwts
-    properties["presentation_submission"] = presentationSubmission
+    val properties: Map<String, Any> = mapOf(
+      "verifiableCredential" to createVpOptions.verifiableCredentialJwts,
+      "presentation_submission" to presentationSubmission
+    )
 
     val vp: VerifiablePresentation? = VerifiablePresentationType.builder()
       .properties(properties)
@@ -309,22 +281,6 @@ public object VerifiablePresentation {
   }
 
   /**
-   * Decodes a verifiable presentation JWT into its header, payload, and signature components.
-   *
-   * @param vpJWT The JWT representation of the verifiable presentation to be decoded.
-   * @return A [DecodedJwt] object containing the decoded components.
-   */
-  public fun decode(vpJWT: VpJwt): DecodedJwt {
-    val (encodedHeader, encodedPayload, encodedSignature) = vpJWT.split('.')
-
-    return DecodedJwt(
-      header = String(Base64.getDecoder().decode(encodedHeader)),
-      payload = String(Base64.getDecoder().decode(encodedPayload)),
-      signature = encodedSignature
-    )
-  }
-
-  /**
    * The selectFrom method is a helper function that helps filter out the verifiable credentials which can not be selected and returns
    * the selectable credentials.
    *
@@ -345,6 +301,10 @@ public object VerifiablePresentation {
       // Fields Processing
       if (inputDescriptor.constraints.fields!!.isNotEmpty()) {
         for (vcJwt: VcJwt in vcJwts) {
+
+          val vc: VerifiableCredentialType =
+            FromJwtConverter.fromJwtVerifiableCredential(JwtVerifiableCredential.fromCompactSerialization(vcJwt))
+
           var fieldMatch = false
 
           for (field: FieldV2 in inputDescriptor.constraints.fields) {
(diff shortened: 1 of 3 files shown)


credentials/src/main/kotlin/SSI.kt
```diff
@@ -270,30 +273,26 @@ public object VerifiablePresentation {
...
-      throw Exception("There are no useable Vcs that correspond to the presentation definition")
-    }
-
-    // TODO change to be more than one VC
-    val vcToUse: VerifiableCredentialType =
-      FromJwtConverter
-        .fromJwtVerifiableCredential(JwtVerifiableCredential.fromCompactSerialization(usableVcJwts.get(0)))
+    val properties: MutableMap<String, Any> = HashMap()
+    properties["verifiableCredential"] = createVpOptions.verifiableCredentialJwts
+    properties["presentation_submission"] = presentationSubmission
```
andresuribe87, line 283: Use `mapOf` instead, so the map is immutable. 

Better yet, create a type for the container, instead of using a `Map<String, Any>`.

credentials/src/main/kotlin/SSI.kt
```diff
@@ -308,24 +307,40 @@ public object VerifiablePresentation {
...
-@Throws(Exception::class)
-private fun selectFrom(presentationDefinition: PresentationDefinitionV2, vcJwts: List<VcJwt>): List<VcJwt> {
-  val selectableCredentials = mutableListOf<VcJwt>()
+  /**
+   * Decodes a verifiable presentation JWT into its header, payload, and signature components.
+   *
+   * @param vpJWT The JWT representation of the verifiable presentation to be decoded.
+   * @return A [DecodedJwt] object containing the decoded components.
+   */
+  public fun decode(vpJWT: VpJwt): DecodedJwt {
```
andresuribe87, line 317: Can this be made private?

credentials/src/main/kotlin/SSI.kt
```diff
@@ -308,24 +307,40 @@ public object VerifiablePresentation {
...
+    val (encodedHeader, encodedPayload, encodedSignature) = vpJWT.split('.')
+
+    return DecodedJwt(
+      header = String(Base64.getDecoder().decode(encodedHeader)),
+      payload = String(Base64.getDecoder().decode(encodedPayload)),
+      signature = encodedSignature
+    )
```
andresuribe87, line 324: Nimbus already has this. Is it possible to remove the `DecodedJwt`? Why is it part of our API? See https://connect2id.com/products/nimbus-jose-jwt/examples/jose-jwt-parsing

credentials/src/main/kotlin/SSI.kt
```diff
@@ -359,21 +374,86 @@ private fun selectFrom(presentationDefinition: PresentationDefinitionV2, vcJwts:
...
+      if (inputDescriptor.constraints.fields!!.isNotEmpty()) {
+        for (field: FieldV2 in inputDescriptor.constraints.fields) {
+          if (field.optional == true) continue // Skip optional fields
 
-  return result
+          var fieldMatch = false
+          for ((vcIndex, vcJwt) in vcJwts.withIndex()) {
+            for (path: String in field.path) {
+              val jsonPathResult = evaluateJsonPath(vcJwt, path)
+              if (jsonPathResult != null) {
```
andresuribe87, line 421: I think this might simplify the code
```suggestion
            for (jsonPathResult: String in field.path.mapNotNull{ evaluateJsonPath(vcJwt, it) }) {
```

credentials/src/main/kotlin/SSI.kt
```diff
@@ -388,4 +468,14 @@ private fun issuerPublicJWK(vcJWT: String, resolver: DIDResolver): JWK {
 
   val publicKeyJWK = JWK.parse(publicKeyJwkMap)
   return publicKeyJWK
+}
+
+private fun evaluateJsonPath(vcJwt: VcJwt, path: String): String? {
+  val vc: VerifiableCredentialType =
```
andresuribe87, line 474: Thi JWT conversion could possibly be done outside the loop that is calling this function.

credentials/src/main/kotlin/model/SSITypes.kt
```diff
@@ -173,3 +175,25 @@ public class FilterV2(
...
+public data class PresentationSubmission(
+  val id: String,
+  val definitionId: String,
+  val descriptorMap: List<DescriptorMap>
+)
+
+/**
+ * Represents descriptor map for a presentation submission.
+ */
+@JsonInclude(JsonInclude.Include.NON_NULL)
```
andresuribe87, line 193: Curious what ths does?

credentials/src/test/kotlin/SSITest.kt
```diff
@@ -108,6 +111,40 @@ class SSITest {
...
+  fun `creates credential status vc with valid vc builder returns vc jwt`() {
+    val credentialSubject = CredentialSubject.builder()
+      .id(URI.create(did))
+      .claims(mutableMapOf<String, Any>().apply { this["firstName"] = "Bobby" })
+      .build()
+
+    val properties: MutableMap<String, Any> = HashMap()
+    properties["statusPurpose"] = "revocation"
+    properties["statusListIndex"] = "94567"
+    properties["statusListCredential"] = "https://example.com/credentials/status/3"
```
andresuribe87, line 124: Is it possible to us `mapOf`?

credentials/src/test/kotlin/SSITest.kt
```diff
@@ -108,6 +111,40 @@ class SSITest {
     assertTrue(VerifiableCredential.verify(vcJwt, SimpleResolver(didDocument)))
   }
 
+  @Test
+  fun `creates credential status vc with valid vc builder returns vc jwt`() {
```
andresuribe87, line 115: Is credential status intended in this PR?

credentials/src/test/kotlin/SSITest.kt
```diff
@@ -124,10 +161,29 @@ class SSITest {
...
+    val header = String(Base64.getDecoder().decode(parts[0]))
+    val payload = String(Base64.getDecoder().decode(parts[1]))
+
+    // Header Checks
+    assertTrue { header.contains("\"alg\":\"") }
+    assertTrue { header.contains("\"typ\":\"JWT\"") }
+
+    // Payload Checks
+    assertTrue { payload.contains("\"iss\":\"") }
+    assertTrue { payload.contains("\"sub\":\"") }
```
andresuribe87, line 186: Seems strange to me to not use the nimbus classes. What's the motivation here?

credentials/src/main/kotlin/SSI.kt
nitro-neal: yup I'll use mapOf

I think the object still needs to be Map<String, Any> because thats what .properties() accepts on the builder

credentials/src/main/kotlin/SSI.kt
nitro-neal: They are not used for now, and really not needed since we have nimbus.

I will remove these

nitro-neal: In the future we could have a nice VcJwtDecoded object with the full VC paylaod already cast for us

credentials/src/main/kotlin/SSI.kt
nitro-neal: Removed

credentials/src/main/kotlin/SSI.kt
nitro-neal: true, this would make it faster in most cases

credentials/src/main/kotlin/model/SSITypes.kt
nitro-neal: if fixed our json serialization so that if something is null it just does not show up in a jsonObject:
```
vp :{
  vc : {...},
  format: null
}

VS

vp :{
  vc : {...}
}
```

credentials/src/test/kotlin/SSITest.kt
nitro-neal: Yea added it for an example that we have the base cred status objects, more robust vc status will come in future prs

credentials/src/test/kotlin/SSITest.kt
nitro-neal: Quick and dirty validation, 
switched to nimbus! 👍

andresuribe87 (approved):

PR #49, opened 2023-10-03
Cred status list

nitro-neal:
# Overview
Cred Credential Status Lists for Revocation (or Suspension)

# Description
This change allows for generation of a status list credential. *NOTE* This is a stateless SDK and it is up to the issuer to set the IDs and host these creds for public retrieval

```
  /**
   * Generates a `StatusList2021Credential` in JWT format for a list of issued credentials.
   *
   * The function creates a status list credential with a specific purpose, e.g., for revocation.
   * The credential indicates the status (like revocation status) of other credentials.
   *
   * @param signOptions The options required for signing the generated status list credential.
   * @param statusListCredentialId The identifier for the status list credential.
   * @param issuer The DID or URI of the entity issuing the status list credential.
   * @param purpose The purpose of the status list (e.g., "revocation").
   * @param issuedCredentialJwts A list of JWTs representing the credentials whose status needs to be represented in the status list.
   * @return A [VcJwt] representing the `StatusList2021Credential` in JWT format.
   * @throws Exception Throws an exception if the generation of the `StatusList2021Credential` fails or any other unexpected error occurs during the generation process.
   */
```

## References
https://www.w3.org/TR/vc-status-list/


credentials/src/test/kotlin/SSICredStatusListTest.kt
```diff
@@ -0,0 +1,277 @@
...
+    }
+
+    assertTrue(
+      exception
+        .message!!.contains("invalid status list index")
+    )
+  }
+
+  @Test
+  fun `validate credential exists in status cred list`() {
```
andresuribe87, line 198: Consider using test parameterization. See https://proandroiddev.com/kotlin-unit-tests-with-parameters-e37aab2b36f6 for an example. This should simplify the test code and make it easier to grok. It would also separate the different behaviors that you're testing, into singular tests.

credentials/src/test/kotlin/SSICredStatusListTest.kt
```diff
@@ -0,0 +1,277 @@
...
+        revocationID,
+        testIssuer,
+        "revocation",
+        listOf(testCred1)
+      )
+    }
+
+    assertTrue(
+      exception
+        .message!!.contains("invalid status list index")
```
andresuribe87, line 193: Would be useful for the message to contain what the status list index value was, and where it was found.

credentials/src/test/kotlin/SSICredStatusListTest.kt
```diff
@@ -0,0 +1,277 @@
...
+        revocationID,
+        testIssuer,
+        "revocation",
+        listOf(testCred1, testCred2)
+      )
+    }
+
+    assertTrue(
+      exception
+        .message!!.contains("duplicate entry found")
```
andresuribe87, line 165: Would be useful to have the message contain which entry was the duplicate entry found, and which credentials it belonged to.

credentials/src/main/kotlin/SSI.kt
```diff
@@ -228,6 +233,220 @@ public object VerifiableCredential {
...
+   * @return A [VcJwt] representing the `StatusList2021Credential` in JWT format.
+   * @throws Exception Throws an exception if the generation of the `StatusList2021Credential` fails or any other unexpected error occurs during the generation process.
+   */
+  @Throws(Exception::class)
+  public fun generateStatusList2021Credential(
+    signOptions: SignOptions,
+    statusListCredentialId: String,
+    issuer: String,
+    purpose: String,
+    issuedCredentialJwts: List<VcJwt>
```
andresuribe87, line 269: Clients that use this SDK may or may not be storing `VcJwt` objects in their DBs. As such, let's use an approach that caters to both. The name BitStringStatusListEntry comes from the [latest draft of the spec](https://w3c.github.io/vc-status-list-2021/#bitstringstatuslistentry). Crucially, this shouldn't be a secured thing, since the SDK is not performing any verification on the data structure (i.e. it should not be a JWT). 
```suggestion
    issuedCredentialStatusListEntries: Iterable<BitStringStatusListEntry >
```

credentials/src/main/kotlin/SSI.kt
```diff
@@ -228,6 +233,220 @@ public object VerifiableCredential {
...
+   * @param issuedCredentialJwts A list of JWTs representing the credentials whose status needs to be represented in the status list.
+   * @return A [VcJwt] representing the `StatusList2021Credential` in JWT format.
+   * @throws Exception Throws an exception if the generation of the `StatusList2021Credential` fails or any other unexpected error occurs during the generation process.
+   */
+  @Throws(Exception::class)
+  public fun generateStatusList2021Credential(
+    signOptions: SignOptions,
+    statusListCredentialId: String,
+    issuer: String,
+    purpose: String,
```
andresuribe87, line 268: `revocation` and `suspension` have specific meaning. I think we should have an enum that encodes that.

credentials/src/main/kotlin/SSI.kt
```diff
@@ -228,6 +233,220 @@ public object VerifiableCredential {
...
+
+  /**
+   * Converts a verifiable credential JWT into a [VerifiableCredentialType] object.
+   *
+   * @param vcJwt The verifiable credential in JWT format to be converted.
+   * @return A [VerifiableCredentialType] object representing the provided verifiable credential JWT.
+   * @throws Exception Throws an exception if the conversion from JWT to [VerifiableCredentialType] fails or any other unexpected error occurs during the conversion process.
+   */
+  @Throws(Exception::class)
+  public fun fromJwt(vcJwt: VcJwt): VerifiableCredentialType {
```
andresuribe87, line 245: Would be better off as an extension function IMHO. See https://kotlinlang.org/docs/coding-conventions.html#extension-functions

credentials/src/main/kotlin/SSI.kt
```diff
@@ -228,6 +233,220 @@ public object VerifiableCredential {
...
+   * The credential indicates the status (like revocation status) of other credentials.
+   *
+   * @param signOptions The options required for signing the generated status list credential.
+   * @param statusListCredentialId The identifier for the status list credential.
+   * @param issuer The DID or URI of the entity issuing the status list credential.
+   * @param purpose The purpose of the status list (e.g., "revocation").
+   * @param issuedCredentialJwts A list of JWTs representing the credentials whose status needs to be represented in the status list.
+   * @return A [VcJwt] representing the `StatusList2021Credential` in JWT format.
+   * @throws Exception Throws an exception if the generation of the `StatusList2021Credential` fails or any other unexpected error occurs during the generation process.
+   */
```
andresuribe87, line 262: Apparently the kotlin coding convention guidance is to not use `@param` and `@return`. See [here](https://kotlinlang.org/docs/coding-conventions.html#documentation-comments)

credentials/src/test/kotlin/SSICredStatusListTest.kt
andresuribe87, line 1: I think adding the example of https://w3c.github.io/vc-status-list-2021/#examples would be important to make sure the implementation is tracking the spec.

credentials/src/main/kotlin/SSI.kt
```diff
@@ -228,6 +233,220 @@ public object VerifiableCredential {
...
+   *
+   * @param credentialToValidateJwt The JWT representing the credential to be validated.
+   * @param statusListCredentialJwt The JWT representing the `StatusList2021Credential`.
+   * @return `true` if the credential's status list index is present in the status list (revoked), `false` otherwise.
+   * @throws Exception Throws an exception if there is a mismatch in status purposes between the two credentials, if required fields are missing, or if any other unexpected error occurs during validation.
+   */
+  @Throws(Exception::class)
+  public fun validateCredentialInStatusList(
+    credentialToValidateJwt: VcJwt,
+    statusListCredentialJwt: VcJwt
```
andresuribe87, line 307: The `credentialToValidateJwt.credentialStatus.statusListCredential` can be dereferenced to obtain the  `statusListCredentialJwt`. Given that, should the `statusListCredentialJwt` parameter be removed from this function?

andresuribe87 (changes requested):

credentials/src/main/kotlin/SSI.kt
nitro-neal: Agree! Added an enum to the SSITypes.ts

credentials/src/main/kotlin/SSI.kt
```diff
@@ -228,6 +233,220 @@ public object VerifiableCredential {
...
+   * 3. Compresses the generated bitstring using the GZIP compression algorithm.
+   * 4. Returns the base64-encoded representation of the compressed bitstring.
+   *
+   * @param statusListIndexes A list of status list index values, represented as strings.
+   * @return A base64-encoded, GZIP-compressed bitstring representation of the provided status list indexes.
+   * @throws Exception Throws an exception if an invalid or duplicate status list index value is found.
+   */
+  private fun bitstringGeneration(statusListIndexes: List<String>): String {
+    val duplicateCheck = mutableSetOf<Int>()
+    val bitSet = BitSet(16 * 1024 * 8) // 1KB = 1024 bytes, 1 byte = 8 bits
```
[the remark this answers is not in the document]
nitro-neal: Yup great catch, I added a check for this

credentials/src/main/kotlin/SSI.kt
andresuribe87: A middle ground may be to let users specify the http client they want to use. This allows easy mocking, and prevents misusing the status list API.

credentials/src/test/kotlin/SSICredStatusListTest.kt
nitro-neal: In doing this I realized that this is using VC 2.0,
Currently we -can- do vc 2.0 with the library we are using in a slightly round about way (JSON LD),
But we are supporting 1.1 in our normal create() method and such

I think we should stick to the examples here which show 1.1 creds -
https://www.w3.org/TR/vc-status-list/

credentials/src/test/kotlin/SSICredStatusListTest.kt
nitro-neal: Happy to do this in a future PR, our tests definitely have a lot of reuse. 
https://github.com/TBD54566975/web5-sdk-kotlin/issues/51

credentials/src/main/kotlin/SSI.kt
nitro-neal: Yup very cool, perfect use of this ✅

nitro-neal:
Made changes in resolved comments 😄

credentials/src/test/kotlin/SSICredStatusListTest.kt
andresuribe87: Oh I was searching for that one, thanks!! Agree that we should stick to 1.1, and wait for 2.0 to be ratified.

nitro-neal:
Make sure there is no conflict with Moe's PR that he has updated - https://github.com/TBD54566975/web5-sdk-kotlin/pull/34

credentials/src/main/kotlin/SSI.kt
nitro-neal: Added this 👍  I think it will work pretty well, if they dont include anything there is a default http client that is created

credentials/src/main/kotlin/SSI.kt
```diff
@@ -228,6 +276,227 @@ public object VerifiableCredential {
...
+   * Validates if a given credential is part of the status list represented by a `StatusList2021Credential`.
+   *
+   * The function takes in a credential representing the one to be validated.
+   * It fetches the status list credential from a URL present in the provided credential.
+   * It then checks if the given credential's status list index is present in the expanded status list derived from the status list credential.
+   */
+  @Throws(Exception::class)
+  public suspend fun validateCredentialInStatusList(
+    credentialToValidate: VerifiableCredentialType,
+    httpClient: HttpClient = defaultHttpClient() // default HTTP client but can be overridden
```
andresuribe87, line 355: HttpClients should be closed in order to free up connection resources. What do you think would be a good approach here?

credentials/src/test/kotlin/SSICredStatusListTest.kt
```diff
@@ -0,0 +1,433 @@
...
+                "type": "BitStringStatusListEntry",
+                "statusPurpose": "revocation",
+                "statusListIndex": "94567",
+                "statusListCredential": "https://example.com/credentials/status/3"
+              },
+              "credentialSubject": {
+                "id": "did:example:6789",
+                "type": "Person"
+              }
+            }"""
```
andresuribe87, line 71: What do you think about putting this in a separate file with JSON?

credentials/src/test/kotlin/SSICredStatusListTest.kt
```diff
@@ -0,0 +1,433 @@
...
+    }
+
+    assertTrue(
+      exception
+        .message!!.contains("duplicate entry found with index: 123")
+    )
+  }
+
+  @Test
+  fun `invalid index value`() {
```
andresuribe87, line 257:
```suggestion
  fun `generating list with negative index value fails`() {
```

credentials/src/test/kotlin/SSICredStatusListTest.kt
```diff
@@ -0,0 +1,433 @@
...
+    }
+
+    assertTrue(
+      exception
+        .message!!.contains("invalid status list index: -1")
+    )
+  }
+
+  @Test
+  fun `invalid value too large for bitset`() {
```
andresuribe87, line 284:
```suggestion
  fun `generating list with large index value fails`() {
```

credentials/src/test/kotlin/SSICredStatusListTest.kt
```diff
@@ -0,0 +1,433 @@
...
+    )
+
+    assertEquals(
+      "H4sIAAAAAAAA/2NgQAESAAPT1/8QAAAA",
+      statusListCredential.credentialSubject.jsonObject["encodedList"] as? String?
+    )
+  }
+
+  @Test
+  fun `fails when duplicate statusListIndex`() {
```
andresuribe87, line 223:
```suggestion
  fun `generating list with duplicate statusListIndex fails`() {
```

credentials/src/test/kotlin/SSICredStatusListTest.kt
```diff
@@ -0,0 +1,433 @@
...
+
+    val revoked2 = VerifiableCredential.validateCredentialInStatusList(testCred2, statusListCredential)
+    assertTrue(revoked2)
+
+    val revoked3 = VerifiableCredential.validateCredentialInStatusList(testCred3, statusListCredential)
+    assertFalse(revoked3)
+  }
+
+  @Test
+  fun `validate credential in status list using async method`() = runBlocking {
```
andresuribe87, line 355:
```suggestion
  fun `validate credential in status list using async method`() = runTest {
```

credentials/src/test/kotlin/SSICredStatusListTest.kt
```diff
@@ -0,0 +1,433 @@
...
+    }
+
+    assertTrue(
+      exception
+        .message!!.contains("invalid status list index: ${Int.MAX_VALUE}, index is larger than the bitset size")
+    )
+  }
+
+  @Test
+  fun `validate credential exists in status cred list`() {
```
andresuribe87, line 311: This tests is the same as below, yes? If so, would it make sense to remove it?

andresuribe87 (approved):
Awesome stuff! Some minor nits for readability.

credentials/src/main/kotlin/SSI.kt
```diff
@@ -228,6 +276,227 @@ public object VerifiableCredential {
...
+  }
+
+  /**
+   * Validates if a given credential is part of the status list represented by a `StatusList2021Credential`.
+   *
+   * The function takes in two credentials: one representing the credential to be validated and the other representing the status list credential.
+   * It checks if the given credential's status list index is present in the expanded status list derived from the status list credential.
+   */
+  @Throws(Exception::class)
+  public fun validateCredentialInStatusList(
```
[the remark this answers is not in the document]
nitro-neal: I kept both public. I'm 90% sure a user can be in a situation in code where they have the status list, but its not hosted anywhere (or behind a firewall or something) and they will need to check if a cred is revoked against it

credentials/src/main/kotlin/SSI.kt
nitro-neal: Updated this!
I handle the closing if its the default client, otherwise if a user defines a client they will be the ones expected to close it

nitro-neal:
closing for this one - https://github.com/TBD54566975/web5-kt/pull/84

PR #50, opened 2023-10-04
Did ION create and resolve

### LINKED ISSUE #4: Implement did:ion Create Operation
Use `create()` method in TypeScript SDK as a reference:

https://github.com/TBD54566975/web5-js/blob/main/packages/dids/src/did-ion.ts#L137

### LINKED ISSUE #5: Implement did:ion Resolution
Use `resolve()` method in TypeScript SDK as a reference:

https://github.com/TBD54566975/web5-js/blob/main/packages/dids/src/did-ion.ts#L429

### LINKED ISSUE #6: Implement did:ion Anchoring
Use `anchor()` method in TypeScript SDK as a reference:

https://github.com/TBD54566975/web5-js/blob/main/packages/dids/src/did-ion.ts#L79

### LINKED ISSUE #7: Implement did:key Create Operation
Use `create()` method in TypeScript SDK as a reference:

https://github.com/TBD54566975/web5-js/blob/main/packages/dids/src/did-key.ts#L86

andresuribe87:
# Overview
Implementes ION creation and resolution. Fixes #5. Fixes #6. And fixes #4 .

# Description
This PR adds:
* Models for sidetree based ion methods which can be serialized to/from json.
* An implementation of a `DIDIonManager`, which can be used to `create` and `resolve` identifiers. 
* The object is configurable so callers can pass custom client engines, and customize the ion URL.
* Changes the interface `DidMethod` to force method specific options, instead of doing runtime casts.

# How Has This Been Tested?
Unit tests that mock the endpoints.
I manually tested using the host `https://ion.tbddev.org` to verify that creation works. An integration test (or e2e test) will need to be added, for which I'll create an issue.

## References
* https://identity.foundation/sidetree/spec/#create


dids/src/main/kotlin/DIDIonManager.kt
```diff
@@ -0,0 +1,283 @@
...
+      Base64URL.encode(
+        Multihash.sum(Multicodec.SHA2_256, canonicalized(createOp.suffixData)).get()?.bytes()
+      ).toString()
+    val initialState = InitialState(
+      suffixData = createOp.suffixData,
+      delta = createOp.delta,
+    )
+    val longFormDIDSegment = didUriSegment(initialState)
+
+    val response: HttpResponse = client.post(operationsEndpoint) {
```
nitro-neal, line 127: is this anchoring on every did we make? Maybe even in testing?

Should we have a flag pass through to use longform only instead of anchoring?

dids/src/main/kotlin/DIDIonManager.kt
```diff
@@ -0,0 +1,283 @@
...
+  }
+
+  /**
+   * Given a [did], returns the [ResolveRepresentationResult], which is specified in https://w3c-ccg.github.io/did-resolution/#did-resolution-result
+   */
+  public suspend fun resolve(did: DID): ResolveDataModelResult {
+    require(did.methodName == "ion")
+
+    val resp = client.get(
+      "https://ion.tbddev.org/identifiers/$did"
```
nitro-neal, line 174: replace with ionHost + "/identifiers/$did"

dids/src/main/kotlin/DIDIonManager.kt
```diff
@@ -0,0 +1,283 @@
...
+    // 1. Encode the public key into the form of a valid JWK.
+    val pkJson = publicKeyJWK.toJSONString()
+
+    // 2. Canonicalize the JWK encoded public key using the implementation’s JSON_CANONICALIZATION_SCHEME.
+    val canonicalized = JsonCanonicalizer(pkJson).encodedUTF8
+
+    // 3. Use the implementation’s HASH_PROTOCOL to Multihash the canonicalized public key to generate the REVEAL_VALUE,
+    val intermediate = Multihash.sum(Multicodec.SHA2_256, canonicalized).get()?.digest!!
+
+    // then Multihash the resulting Multihash value again using the implementation’s HASH_PROTOCOL to produce
```
nitro-neal, line 254: put // 4. to stick with comment format

dids/src/main/kotlin/DIDIonManager.kt
```diff
@@ -0,0 +1,283 @@
...
+   * Default companion object for creating a DIDIonManager with a default configuration.
+   */
+  public companion object Default : DIDIonManager(DIDIonConfiguration())
+}
+
+/**
+ * Metadata related to the creation of a DID (Decentralized Identifier) on the Sidetree protocol.
+ *
+ * @property createOperation The Sidetree create operation used to create the DID.
+ * @property shortFormDID The short-form DID representing the DID created.
```
nitro-neal, line 270: reverse property order to math actual function

dids/src/main/kotlin/ion/model/Models.kt
```diff
@@ -0,0 +1,205 @@
...
+  public val id: String,
+  public val type: String,
+  public val serviceEndpoint: String
+)
+
+/**
+ * Represents a public key in the ION document.
+ *
+ * @property id The public key ID.
+ * @property type The public key type.
```
nitro-neal, line 41: controller missing

dids/src/main/kotlin/ion/model/Models.kt
```diff
@@ -0,0 +1,205 @@
...
+ * @property publicKeyJWK The public key JWK (JSON Web Key).
+ * @property purposes List of public key purposes.
+ */
+@Serializable
+public data class PublicKey(
+  public val id: String,
+  public val type: String,
+  public val controller: String? = null,
+  @SerialName("publicKeyJwk") public val publicKeyJWK: JsonWebKey? = null,
```
nitro-neal, line 50: I think throughout the code we have Jwt

So yea happy to change to JWT if its the suggested way, but yea needs to be consistent (maybe in a new PR)

dids/src/main/kotlin/ion/model/Models.kt
```diff
@@ -0,0 +1,205 @@
...
+ * @property x5tS256 X.509 Certificate SHA-256 Thumbprint.
+ * @property x5uHeaderParam Custom X.509 URL Header Parameter.
+ * @property x5cHeaderParam List of Custom X.509 Certificate Chain Header Parameters.
+ * @property crv Curve.
+ * @property x X Coordinate.
+ * @property y Y Coordinate.
+ * @property d D.
+ */
+@Serializable
+public data class JsonWebKey(
```
nitro-neal, line 74: There is no nimbus class for this?

dids/src/test/kotlin/DIDIonTest.kt
```diff
@@ -0,0 +1,118 @@
...
+  @Test
+  fun createWithCustom() = runTest {
+    val verificationKey = ECKeyGenerator(Curve.SECP256K1)
+      .keyUse(KeyUse.SIGNATURE)
+      .keyID(UUID.randomUUID().toString())
+      .provider(BouncyCastleProviderSingleton.getInstance() as Provider)
+      .generate()
+      .toPublicJWK()
+      .toJsonWebKey()
+    val c = DIDIonManager {
```
nitro-neal, line 41:
```suggestion
    val didIonManager = DIDIonManager {
```

nitro-neal (dismissed):

nitro-neal:
Actually looking at Moes I think it would be much easier if he merges first, and then you build did ion extending the stuff he has

dids/src/main/kotlin/ion/model/Models.kt
andresuribe87: There is, BUT, I could get a class like the example below to serializae using `kotlinx.serialization`

```kt
@Serializable
public data class Foo(
  public val jwk : JWK
)
```


It failed when calling `Json.decodeFromString(fooStringJson)`

dids/src/main/kotlin/ion/model/Models.kt
andresuribe87: Just making sure. JWT is a Json Web Token. JWK is a Json Web Key. They're different things.

Assuming you meant that we should rename the field to `jwk`? If so, this is implementing the sidetree spec, which contains the names already. It needs to be `publicKeyJwk`. So leaving as is.

dids/src/main/kotlin/ion/model/Models.kt
andresuribe87: Removed the properties and pointed to the spec, which is the source of truth.

andresuribe87:
@nitro-neal I took a pass and incorporated the new mainline.

bradleydwyer:
What was the motivation behind choosing to adopt coroutines in this work?

bradleydwyer:
Same question for the experimental Kotlin serialization APIs

andresuribe87:
> What was the motivation behind choosing to adopt coroutines in this work?

Operations related to ion dis require network calls, and we chose ktor for network calls. Ktor is built on top of coroutines. Ktor seemed like a sensible choice for multiplatform support.

andresuribe87:
> Same question for the experimental Kotlin serialization APIs

Kotlin serialization is not experimental. Some parts of the API are experimental, meaning that the API might change in the future. That's should be an implementation detail for users of the SDK. 

The advantage is that the code is more succinct, and we get CBOR almost for free.

This isn't a final decision. We can always change. Are there use cases you are targeting that would be hard  to meet with the current choices?

bradleydwyer:
> > What was the motivation behind choosing to adopt coroutines in this work?
> 
> Operations related to ion dis require network calls, and we chose ktor for network calls. Ktor is built on top of coroutines. Ktor was a sensible choice for multiplatform support.

We should be intentional here in the decision. coroutines are a PITA in practice. Moreover once they are in they're force spread everywhere up the call stack. We should also be aware of how they surface to Java callers.

bradleydwyer:
All of the `Exception` references in this PR (and elsewhere in the lib) really should be RuntimeException. Even through in Kotlin Exception isn't a checked exception, it will be when invoked from Java.

bradleydwyer:
> All of the `Exception` references in this PR (and elsewhere in the lib) really should be `RuntimeException`. Even though in Kotlin `Exception` isn't a checked exception, it will be when invoked from Java.

In addition to this, generic `throws Exception` instances should be replaced by more specific exception types, rather than relying on callers to parse the message.

dids/src/main/kotlin/ion/model/Models.kt
mistermoe: Let's stick with using nimbusds `JWK`to remain consistent with the rest of the codebase

mistermoe:
* let's stick with a synchronous api surface for the initial release of these libs. If we introduce async later on, I think it should be in addition to the sync api not a replacement of. For now I think we should focus on ensuring everything works correctly 
* we're already using Jackson to serialize / deserialize json and it also provides cbor support. No need to introduce ktor's serialization package atm
* if you have a chance, try using DidIon in a Java project

bradleydwyer:
> * let's stick with a synchronous api surface for the initial release of these libs. If we introduce async later on, I think it should be in addition to the sync api not a replacement of. For now I think we should focus on ensuring everything works correctly

This is getting to what I mean by intentional. We have to make a clear decision on the API surface of this library and whether or not we expose async APIs, synchronous APIs or both; where we surface async APIs we should consider the implications of the style of async API we build as it impacts on users the same way the selection of ktor as a http client drove the use of coroutines in this library.

> * we're already using Jackson to serialize / deserialize json and it also provides cbor support. No need to introduce ktor's serialization package atm

it's kotlin's serialization lib, but again decisions should be intentional. these two paths are not identical and there are tradeoffs on both sides.

andresuribe87:
> > * let's stick with a synchronous api surface for the initial release of these libs. If we introduce async later on, I think it should be in addition to the sync api not a replacement of. For now I think we should focus on ensuring everything works correctly
> 
> This is getting to what I mean by intentional. We have to make a clear decision on the API surface of this library and whether or not we expose async APIs, synchronous APIs or both; where we surface async APIs we should consider the implications of the style of async API we build as it impacts on users the same way the selection of ktor as a http client drove the use of coroutines in this library.

KTOR selection was intentional when I considered:
* Multiplatform. 
Coroutines made sense to me considering that Android code that has to deal with threads is error prone (from my Android experience back in the day) and may block the main thread.

That said, I don't have any preferences at all. As such, I'll make changes so users of the library can actually put in whatever http library they want.

> 
> > * we're already using Jackson to serialize / deserialize json and it also provides cbor support. No need to introduce ktor's serialization package atm
> 
> it's kotlin's serialization lib, but again decisions should be intentional. these two paths are not identical and there are tradeoffs on both sides.

When I drafted this PR, there wasn't any jackson use for json serialization. Did I miss it? 
AFAICT the jackson lib decision was made for `tbdex-kt`. This is a separate repository, so it would warrant a separate set of considerations IMO.
Changing right now just for consistency, but it's something to consider in the future.

andresuribe87 (commented):
@mistermoe ptal

nitro-neal (approved):

andresuribe87:
@mistermoe below is a snippet from what creation looks like in java

```java
    @Test public void didIonGeneration() {
        KeyManager km = new InMemoryKeyManager();
        DidIonHandle did = DidIonManager.Default.create(km, new CreateDidIonOptions());
        assertTrue(did.getUri().startsWith("did:ion"));

        assert did.getCreationMetadata() != null;
        assertTrue(did.getCreationMetadata().getLongFormDid().startsWith(did.getUri()));
    }
```

mistermoe:
@andresuribe87 i noticed we have a transitive dependency that uses `slf4j`. do you happen to know which dep it is? ran the following:

```kotlin
package web5.sdk.dids

import web5.sdk.crypto.InMemoryKeyManager

public fun main() {
  val km = InMemoryKeyManager()
  val did = DidIonManager.create(km)

  DidIonManager.resolve(did.uri)
}
```

saw:

```
SLF4J: No SLF4J providers were found.
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See https://www.slf4j.org/codes.html#noProviders for further details.
```

mistermoe:
[didlint](https://didlint.ownyourdata.eu/validate?did=did:ion:EiD2QfW14vmqLto_yztr3OLL_qQKXvVCagsaDgIyA8fgjA:eyJkZWx0YSI6eyJwYXRjaGVzIjpbeyJhY3Rpb24iOiJyZXBsYWNlIiwiZG9jdW1lbnQiOnsicHVibGljS2V5cyI6W3siaWQiOiIyY2RjOGFjZi1mNDkxLTRkM2YtYjE5Yi05MmI4NjU5N2JiYWEiLCJwdWJsaWNLZXlKd2siOnsiYWxnIjoiRVMyNTZLIiwiY3J2Ijoic2VjcDI1NmsxIiwia2lkIjoiVHB0ejBaNTdST1VYaExjWV85OE9OcFQ2QXRwQ0lVdkJtcDVCX05Rd3hJNCIsImt0eSI6IkVDIiwidXNlIjoic2lnIiwieCI6IkFwZER3aDNyS2dOMGlDZy1VaU8tU3NCTzl3NnBlWXpiOWd1YkN3aWZCcTAiLCJ5IjoibnVkU24wTFY2MThsM0ZHNVQ0QWRDV2tPdXJaX1c3S19CNnVxb2E2OG5hVSJ9LCJwdXJwb3NlcyI6WyJhdXRoZW50aWNhdGlvbiJdLCJ0eXBlIjoiSnNvbldlYktleTIwMjAifV0sInNlcnZpY2VzIjpbXX19XSwidXBkYXRlQ29tbWl0bWVudCI6IkVpRHVDb0ctSVRNSzRBYk15TzNLTmw0eEd5cHJPeTBvdHI1MzduRDBDN0VzdFEifSwic3VmZml4RGF0YSI6eyJkZWx0YUhhc2giOiJFaUNYZTd2NjVra0I4NWdEMjgzOXJMVlpZYnVZWDU3Xy1GNlR5U1NWcThyQURRIiwicmVjb3ZlcnlDb21taXRtZW50IjoiRWlCX3FuWTBScDNaOGJaNGdVdlh3clhIRHJkTGNqOXVQdWFDcWFnSHFtYnBHQSJ9fQ) states there's some issues with the generated did document

crypto/src/main/kotlin/web5/sdk/crypto/InMemoryKeyManager.kt
```diff
@@ -55,7 +56,7 @@ public class InMemoryKeyManager : KeyManager {
   override fun getPublicKey(keyAlias: String): JWK {
     // TODO: decide whether to return null or throw an exception
     val privateKey = getPrivateKey(keyAlias)
-    return Crypto.computePublicKey(privateKey)
+    return privateKey.toPublicJWK()
```
mistermoe, line 59: would prefer that we use `computePublicKey` since it works.

crypto/src/main/kotlin/web5/sdk/crypto/InMemoryKeyManager.kt
```diff
@@ -85,4 +86,19 @@ public class InMemoryKeyManager : KeyManager {
...
+  public fun import(jwk: JWK): String {
+    require(jwk.isPrivate) {
+      "Importing a non-private key is not permitted"
+    }
+    var kid = jwk.keyID
+    if (kid.isNullOrEmpty()) {
+      kid = UUID.randomUUID().toString()
+    }
+    keyStore.putIfAbsent(kid, jwk)
+    return kid
```
mistermoe, line 102: we should compute the JWK thumbrint if `keyID` is not present

dids/build.gradle.kts
```diff
@@ -13,15 +13,32 @@ repositories {
...
-  implementation("com.google.crypto.tink:tink:1.10.0")
-  implementation("decentralized-identity:did-common-java:1.9.0")
+  implementation("org.erwinkok.multiformat:multiformat:1.1.0")
+  implementation("org.erwinkok.result:result-monad:1.4.0")
+
+  implementation("com.fasterxml.jackson.module:jackson-module-kotlin:2.8.+")
+
+  implementation("io.ktor:ktor-client-core:$ktor_version")
+  implementation("io.ktor:ktor-client-cio:$ktor_version")
+  implementation("io.ktor:ktor-client-content-negotiation:$ktor_version")
```
mistermoe, line 37: curious why we need to bring this in?

dids/src/main/kotlin/web5/sdk/dids/DidIonManager.kt
```diff
@@ -0,0 +1,353 @@
...
+   * Note: [options] must be of type [CreateDidIonOptions].
+   * @throws [ResolutionException] When there is an error after resolution.
+   * @throws [InvalidStatusException] When any of the network requests return an invalid HTTP status code.
+   * @see [DidMethod.create] for details of each parameter.
+   */
+  override fun create(keyManager: KeyManager, options: CreateDidIonOptions?): DidIonHandle {
+    val (createOp, keys) = createOperation(keyManager, options)
+
+    val shortFormDidSegment =
+      Base64URL.encode(
```
mistermoe, line 127: any reason we're not using `Convert`?

dids/src/main/kotlin/web5/sdk/dids/DidIonManager.kt
```diff
@@ -0,0 +1,353 @@
...
+    if (!isBase64UrlString(id)) {
+      throw IllegalArgumentException("verification method id \"$id\" is not base 64 url charset")
+    }
+
+    if (id.length > maxVerificationMethodIdLength) {
+      throw IllegalArgumentException(
+        "verification method id \"$id\" exceeds max allowed length of $maxVerificationMethodIdLength"
+      )
+    }
+  }
```
mistermoe, line 265: is this a sidetree requirement? the did core spec states otherwise

image


with the option of an id simply being `#fragment`.

dids/src/main/kotlin/web5/sdk/dids/DidIonManager.kt
```diff
@@ -0,0 +1,353 @@
...
+    val canonicalized = JsonCanonicalizer(jsonString).encodedUTF8
+    val deltaHash = Multihash.sum(Multicodec.SHA2_256, canonicalized).get()?.bytes()
+    return OperationSuffixDataObject(
+      deltaHash = Base64URL.encode(deltaHash).toString(),
+      recoveryCommitment = recoveryCommitment
+    )
+  }
+
+  private fun publicKeyCommitment(publicKeyJWK: JWK): Commitment {
+    require(!publicKeyJWK.isPrivate)
```
mistermoe, line 285: missing exception message

dids/src/main/kotlin/web5/sdk/dids/DidIonManager.kt
```diff
@@ -0,0 +1,353 @@
...
+    // 1. Encode the public key into the form of a valid JWK.
+    val pkJson = publicKeyJWK.toJSONString()
+
+    // 2. Canonicalize the JWK encoded public key using the implementation’s JSON_CANONICALIZATION_SCHEME.
+    val canonicalized = JsonCanonicalizer(pkJson).encodedUTF8
+
+    // 3. Use the implementation’s HASH_PROTOCOL to Multihash the canonicalized public key to generate the REVEAL_VALUE,
+    val intermediate = Multihash.sum(Multicodec.SHA2_256, canonicalized).get()?.digest!!
```
mistermoe, line 293: handle `digest!!`

dids/src/main/kotlin/web5/sdk/dids/DidIonManager.kt
```diff
@@ -0,0 +1,353 @@
...
+
+    // then Multihash the resulting Multihash value again using the implementation’s HASH_PROTOCOL to produce
+    // the public key commitment.
+    val hashOfHash = Multihash.sum(Multicodec.SHA2_256, intermediate).get()?.bytes()
```
mistermoe, line 297: what if this is null?

dids/src/main/kotlin/web5/sdk/dids/DidIonManager.kt
```diff
@@ -0,0 +1,353 @@
...
+  /**
+   * Default companion object for creating a [DidIonManager] with a default configuration.
+   */
+  public companion object Default : DidIonManager(DidIonConfiguration())
+}
+
+/**
+ * Represents an HTTP response where the status code is outside the range considered success.
+ */
+public class InvalidStatusException(s: String) : Exception(s)
```
mistermoe, line 310: extends `RuntimeException`. Also given that it's a custom exception, would be nice to include http status code

dids/src/main/kotlin/web5/sdk/dids/DidIonManager.kt
```diff
@@ -0,0 +1,353 @@
...
+
+/** Wraps an exception during resolution where the [DidResolutionMetadata.error] is not empty. */
+public class ResolutionException(s: String) : Exception(s)
```
mistermoe, line 313: so does resolve throw an exception?

dids/src/main/kotlin/web5/sdk/dids/DidResolutionResult.kt
```diff
@@ -64,6 +66,16 @@ public class DidDocumentMetadata(
...
+)
+
+/**
+ * Metadata about the did method.
+ */
+public class MetadataMethod(
+  public val published: Boolean,
+  public val recoveryCommitment: String? = null,
+  public val updateCommitment: String? = null,
 )
```
mistermoe, line 81: is this specific to ion?

dids/src/main/kotlin/web5/sdk/dids/ion/model/Models.kt
```diff
@@ -0,0 +1,171 @@
...
+  @JsonDeserialize(using = JacksonJWK.Deserializer::class)
+  public val publicKeyJwk: JWK? = null,
+  public val purposes: List<PublicKeyPurpose> = emptyList()
+)
+
+/**
+ * JacksonJWK is a utility class that facilitates serialization for [JWK] types, so that it's easy to integrate with any
+ * class that is meant to be serialized to/from JSON.
+ */
+public class JacksonJWK {
```
mistermoe, line 60: what was the rationale for putting this class here?

dids/src/main/kotlin/web5/sdk/dids/ion/model/Models.kt
mistermoe, line 1: curious about why `DidIonManager` isn't in the `ion` package but just the models are

dids/src/main/kotlin/web5/sdk/dids/ion/model/Models.kt
```diff
@@ -0,0 +1,171 @@
...
+public data class AddServicesAction(
+  public val services: List<Service> = emptyList()
+) : PatchAction()
+
+/**
+ * Represents a "replace" patch action in the ION document.
+ *
+ * @property document The document to replace.
+ */
+public data class ReplaceAction(
```
mistermoe, line 123: would it be possible to include links to relevant sections of the sidetree spec for each of these data classes?

dids/src/test/kotlin/web5/sdk/dids/DidIonTest.kt
```diff
@@ -0,0 +1,144 @@
...
+  @Test
+  fun createWithCustom() {
+    val keyManager = InMemoryKeyManager()
+    val verificationKey = readKey("src/test/resources/verification_jwk.json")
+    val updateKey = readKey("src/test/resources/update_jwk.json")
+    val recoveryKey = readKey("src/test/resources/recovery_jwk.json")
+    val manager = DidIonManager {
+      ionHost = "madeuphost"
+      engine = mockEngine()
+    }
```
mistermoe, line 71: how does using this end up looking in java?

dids/src/main/kotlin/web5/sdk/dids/DidIonManager.kt
```diff
@@ -0,0 +1,353 @@
...
+    CIO.create {}
+  } else {
+    configuration.engine!!
+  }
+
+  private val client = HttpClient(engine) {
+    install(ContentNegotiation) {
+      jackson { mapper }
+    }
+  }
```
mistermoe, line 110: this implies 1 http client and engine per instantiation?

dids/src/main/kotlin/web5/sdk/dids/DidIonManager.kt
```diff
@@ -0,0 +1,353 @@
...
+import io.ktor.client.request.get
+import io.ktor.client.request.post
+import io.ktor.client.request.setBody
+import io.ktor.client.statement.HttpResponse
+import io.ktor.client.statement.bodyAsText
+import io.ktor.http.ContentType
+import io.ktor.http.contentType
+import io.ktor.http.isSuccess
+import io.ktor.serialization.jackson.jackson
+import kotlinx.coroutines.runBlocking
```
mistermoe, line 22: do we need `coroutine` packages?

dids/src/main/kotlin/web5/sdk/dids/DidIonManager.kt
```diff
@@ -0,0 +1,353 @@
...
+  private val mapper = jacksonObjectMapper()
+
+  private val operationsEndpoint = configuration.ionHost + operationsPath
+  private val identifiersEndpoint = configuration.ionHost + identifiersPath
+
+  private val engine: HttpClientEngine = if (configuration.engine == null) {
+    CIO.create {}
+  } else {
+    configuration.engine!!
+  }
```
mistermoe, line 104:
```suggestion
private val engine: HttpClientEngine = configuration.engine ?: CIO.create {}
```

dids/src/main/kotlin/web5/sdk/dids/DidIonManager.kt
```diff
@@ -0,0 +1,353 @@
...
+          longFormDid,
+          opBody,
+          keys
+        )
+      )
+    }
+    throw InvalidStatusException("received error response '$opBody'")
+  }
+
+  private inline fun <reified T> canonicalized(data: T): ByteArray {
```
mistermoe, line 173: nearly impossible to trace down exceptions when using `inline` functions. Also `reified T` does not work well with Java at all. you're going to end up with just `Object`

dids/src/main/kotlin/web5/sdk/dids/DidIonManager.kt
```diff
@@ -0,0 +1,353 @@
...
+  }
+
+  private fun createOperation(keyManager: KeyManager, options: CreateDidIonOptions?)
+    : Pair<SidetreeCreateOperation, KeyAliases> {
+    val updatePublicJWK: JWK = if (options?.updatePublicJWK == null) {
+      val alias = keyManager.generatePrivateKey(JWSAlgorithm.ES256K, Curve.SECP256K1)
+      keyManager.getPublicKey(alias)
+    } else {
+      options.updatePublicJWK
+    }
```
mistermoe, line 207:
```suggestion
    val updatePublicJWK: JWK = options?.updatePublicJWK ?: keyManager.getPublicKey(
      keyManager.generatePrivateKey(JWSAlgorithm.ES256K)
    )
```

dids/src/main/kotlin/web5/sdk/dids/DidIonManager.kt
```diff
@@ -0,0 +1,353 @@
...
+    val publicKeyCommitment: String = publicKeyCommitment(updatePublicJWK)
+
+    val verificationMethodId = options?.verificationMethodId ?: UUID.randomUUID().toString()
+    validateVerificationMethodId(verificationMethodId)
```
mistermoe, line 211:
```suggestion
    val verificationMethodId = when (options?.verificationMethodId) {
      null -> UUID.randomUUID().toString()
      else -> {
        validateVerificationMethodId(options.verificationMethodId)
        options.verificationMethodId
      }
    }
```

if you want to prevent validating something you just created

dids/src/main/kotlin/web5/sdk/dids/DidIonManager.kt
```diff
@@ -0,0 +1,353 @@
...
+    val verificationPublicKey = if (options?.verificationPublicKey == null) {
+      val alias = keyManager.generatePrivateKey(JWSAlgorithm.ES256K, Curve.SECP256K1)
```
mistermoe, line 213:
```suggestion
      val alias = keyManager.generatePrivateKey(JWSAlgorithm.ES256K)
```

dids/src/main/kotlin/web5/sdk/dids/DidIonManager.kt
```diff
@@ -0,0 +1,353 @@
...
+      options.verificationPublicKey
+    }
+    val patches = listOf(ReplaceAction(Document(listOf(verificationPublicKey))))
+    val createOperationDelta = Delta(
+      patches = patches,
+      updateCommitment = publicKeyCommitment
+    )
+
+    val recoveryPublicJWK = if (options?.recoveryPublicJWK == null) {
+      val alias = keyManager.generatePrivateKey(JWSAlgorithm.ES256K, Curve.SECP256K1)
```
mistermoe, line 231:
```suggestion
      val alias = keyManager.generatePrivateKey(JWSAlgorithm.ES256K)
```

dids/src/main/kotlin/web5/sdk/dids/DidIonManager.kt
```diff
@@ -0,0 +1,353 @@
...
+      SidetreeCreateOperation(
+        type = "create",
+        suffixData = operation,
+        delta = createOperationDelta,
+      ),
+      KeyAliases(
+        updateKeyAlias = updatePublicJWK.keyID,
+        verificationKeyAlias = verificationPublicKey.publicKeyJwk!!.keyID,
+        recoveryKeyAlias = recoveryPublicJWK.keyID
+      )
```
mistermoe, line 251: this would be unnecessary if we take the jwk thumbprint route right?

dids/src/main/kotlin/web5/sdk/dids/DidIonManager.kt
```diff
@@ -0,0 +1,353 @@
...
+
+    if (id.length > maxVerificationMethodIdLength) {
+      throw IllegalArgumentException(
+        "verification method id \"$id\" exceeds max allowed length of $maxVerificationMethodIdLength"
+      )
+    }
+  }
+
+  private fun isBase64UrlString(input: String?): Boolean {
+    val regex = "^[A-Za-z0-9_-]+$".toRegex()
```
mistermoe, line 268: can make this a const

dids/src/main/kotlin/web5/sdk/dids/DidIonManager.kt
```diff
@@ -0,0 +1,353 @@
...
+ * @property longFormDid The long-form DID representing the DID created.
+ * @property operationsResponseBody The response body received after submitting the create operation.
+ */
+public data class IonCreationMetadata(
+  public val createOperation: SidetreeCreateOperation,
+  public val shortFormDid: String,
+  public val longFormDid: String,
+  public val operationsResponseBody: String,
+  public val keyAliases: KeyAliases,
+) : CreationMetadata
```
mistermoe, line 353: what's the deciding factor as to whether something is considered a model or not?

dids/src/main/kotlin/DIDIonManager.kt
mistermoe: weird

andresuribe87 (commented):
thanks for taking a look, all feedback has been addressed now.

dids/src/main/kotlin/web5/sdk/dids/DidIonManager.kt
mistermoe: oh right good call! yep yep

dids/src/main/kotlin/web5/sdk/dids/DidIonManager.kt
mistermoe: no idea. tbh i've never used `ktor` before. i always pick up [okhttp](https://github.com/square/okhttp) because i like how simple it is

mistermoe:
@andresuribe87 it pains me to say this to a primarily Go dev 🙈: 

Convention in this repo is to use camel case instead of screaming case (idk what it's actually called) e.g. `Did` vs `DID`, `Jwk` vs `JWK`. noticed you used both. can you switch uses of screaming -> camel? feel free to leave internal use as screaming (e.g. variable names in function bodies). just want to make sure api surface is consistent

dids/src/main/kotlin/web5/sdk/dids/DidIonManager.kt
```diff
@@ -0,0 +1,356 @@
...
+ * @property uri The URI of the "did:ion" which conforms to the DID standard.
+ * @property keyManager A [KeyManager] instance utilized to manage the cryptographic keys associated with the DID.
+ * @property creationMetadata Metadata related to the creation of a DID. Useful for debugging purposes.
+ *
+ * ### Usage Example:
+ * ```kotlin
+ * val keyManager = InMemoryKeyManager()
+ * val did = DidKey("did:key:example", keyManager)
+ * ```
+ */
```
mistermoe, line 80: update example to show `DidIon` instead of `DidKey`

andresuribe87: Done.

andresuribe87:
> @andresuribe87 it pains me to say this to a primarily Go dev 🙈:
> 
> Convention in this repo is to use camel case instead of screaming case (idk what it's actually called) e.g. `Did` vs `DID`, `Jwk` vs `JWK`. noticed you used both. can you switch uses of screaming -> camel? feel free to leave internal use as screaming (e.g. variable names in function bodies). just want to make sure api surface is consistent

Ah!! I actually had done this, but missed a bunch of places. Now it's done for real. 

I also added an `examples.md` doc, which is references from the README.

dids/src/main/kotlin/web5/sdk/dids/DidIonManager.kt
```diff
@@ -0,0 +1,356 @@
...
+
+    return Pair(
+      SidetreeCreateOperation(
+        type = "create",
+        suffixData = operation,
+        delta = createOperationDelta,
+      ),
+      KeyAliases(
+        updateKeyAlias = updatePublicJwk.keyID,
+        verificationKeyAlias = verificationPublicKey.publicKeyJwk!!.keyID,
```
bradleydwyer, line 251: Please don't use !!

bradleydwyer:
re: https://github.com/TBD54566975/web5-kt/pull/50#issuecomment-1755776229

> KTOR selection was intentional when I considered:

That's not what I meant by intentional. I'm referring to public API design for the SDKs and consistency in implementation (although ktor classes have leaked into the public API in PR 84 - so maybe it is both).

> Coroutines made sense to me considering that Android code that has to deal with threads is error prone (from my Android experience back in the day) and may block the main thread.

Sure but there are many ways to achieve this. In my experience coroutines are error prone and have different behaviour depending on the execution context (e.g. being too clever with thread pools based on underlying hardware).  We may end up using coroutines broadly (I think it is likely) but we should make that call at a library level (matching for tbdex-kt) and not in a single PR – and I definitely think we should have codified appropriate usage patterns first.

I have a general concern of design divergence between everyone that has implemented towards this SDK. It's not a comment against the individuals, but I really think we should consider what the API surface looks like across all the major functional areas for items like the above (async design), instantiation and configuration, types, exceptions vs result types (i.e. functional), validation semantics. We were pressed for time during this so it's not surprising that this is the case.

> This is a separate repository, so it would warrant a separate set of considerations IMO.

I mostly disagree - but I do understand the point. One of the intents of the SDKs was ideally a consistent view of Web5/DIDs/VCs/tbDEX across languages. Consistency in per-language SDK design and implementation seems a foundational part of that.

andresuribe87:
Thanks for the comments @bradleydwyer . Please see below

> re: [#50 (comment)](https://github.com/TBD54566975/web5-kt/pull/50#issuecomment-1755776229)
> 
> > KTOR selection was intentional when I considered:
> 
> That's not what I meant by intentional. I'm referring to public API design for the SDKs and consistency in implementation (although ktor classes have leaked into the public API in PR 84 - so maybe it is both).

In this PR, the public API design only exposes the HttpEngine as configurable (which is absolutely optional and mostly for testing purposes). I considered making the client an interface, but that seemed like a heavy handed approach without a need yet (see https://github.com/TBD54566975/web5-kt/pull/86 for an example of what I mean). If there is a need to support multiple http client frameworks , I'm happy to go that route. I would discourage this though, unless absolutely necessary. Do you have a use case for supporting different types of HTTP client implementations?

Note that this PR needs to have an http client because it's making an HTTP call. There are not examples in this repo that make any network requests right now. Is the implementation consistency you're referring to meant to be across repositories?

> 
> > Coroutines made sense to me considering that Android code that has to deal with threads is error prone (from my Android experience back in the day) and may block the main thread.
> 
> Sure but there are many ways to achieve this. In my experience coroutines are error prone and have different behaviour depending on the execution context (e.g. being too clever with thread pools based on underlying hardware). We may end up using coroutines broadly (I think it is possible) but we should make that call at a library level (matching for tbdex-kt) and not in a single PR – and I definitely think we should have codified appropriate usage patterns first and understand that there are consequences of their use.
> 
> I have a general concern of design divergence between everyone that has implemented towards this SDK. It's not a comment against the individuals, but I really think we should consider what the API surface looks like across all the major functional areas for items like the above (async design), instantiation and configuration, types, exceptions vs result types (i.e. functional), validation semantics. We were pressed for time during this so it's not surprising that this is the case.
> 

Would you mind making a concrete recommendation in this PR so that the API doesn't diverge from what exists? For each of the areas, specifially:
* async - This PR is not introducing any async design. The implementation is now sync, in line with what `DidKey` is doing. 
* instantiation and configuration - `DidKey` doesn't need to do any networking, and is generally very simple. That is, it doesn't required any configuration. `DidIonManager` introduces configuration because it needs to do networking. Can you elaborate on what the divergence is, and a recommended approach?
* Types - I'm not sure what you're referring to, can you clarify?
* Exceptions - All exceptions are being typed in the same way in which they are typed elsewhere in the code. Is there an example you're referring to?
* Result types - Any example you can reference?
* Validation semantics - I'm not sure what you mean by this, can you share an example?

> > This is a separate repository, so it would warrant a separate set of considerations IMO.
> 
> I mostly disagree - but I do understand the point. One of the intents of the SDKs was ideally a consistent view of Web5/DIDs/VCs/tbDEX across languages. Consistency in per-language SDK design and implementation seems a foundational part of that.

Acknowledged. Agree to disagree. That said, I'm still unsure what you're asking for here. Can you point to the specific SDK design choice that's inconsistent here? Alternatively, do you have concrete suggestions on moving this forward?

bradleydwyer (approved):

PR #64, opened 2023-10-09
Fix key type bug in Secp256k1

tomdaffurn:
# Overview
Fix some copy + paste bugs in Secp256k1

### DIFF
--- a/crypto/build.gradle.kts
+++ b/crypto/build.gradle.kts
@@ -15,6 +15,7 @@ dependencies {
   implementation("org.bouncycastle:bcprov-jdk15on:1.70")
   implementation("org.bouncycastle:bcpkix-jdk15on:1.70")
   implementation(project(":common"))
+  testImplementation(kotlin("test"))
 }
 
 tasks.test {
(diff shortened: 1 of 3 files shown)


crypto/build.gradle.kts
```diff
@@ -15,6 +15,7 @@ dependencies {
   implementation("org.bouncycastle:bcprov-jdk15on:1.70")
   implementation("org.bouncycastle:bcpkix-jdk15on:1.70")
   implementation(project(":common"))
+  testImplementation(kotlin("test"))
```
mistermoe, line 18: mostly unrelated to this PR (esp. because i'm using this in `dids` as well), should we use junit5 instead?

mistermoe (approved):
🏅

crypto/src/test/kotlin/web5/sdk/crypto/Secp256k1Test.kt
```diff
@@ -0,0 +1,35 @@
+package web5.sdk.crypto
+
+import com.nimbusds.jose.JWSAlgorithm
+import com.nimbusds.jose.jwk.KeyType
+import com.nimbusds.jose.jwk.KeyUse
+import org.junit.jupiter.api.Assertions.*
```
phoebe-lew, line 6: i think detekt usually chokes on star imports

phoebe-lew (approved):

PR #63, opened 2023-10-09
Update `credentials` package

mistermoe:
TODO: Fill out

PR #65, opened 2023-10-09
AWS Key Manager

tomdaffurn:
# Overview
- Implement a KeyManager that uses AWS KMS
- Improve the KeyManager interface to allow mapping of DIDs to aliases

### DIFF
--- a/crypto/build.gradle.kts
+++ b/crypto/build.gradle.kts
@@ -15,6 +15,10 @@ dependencies {
   implementation("org.bouncycastle:bcprov-jdk15on:1.70")
   implementation("org.bouncycastle:bcpkix-jdk15on:1.70")
   implementation(project(":common"))
+
+  implementation("com.amazonaws:aws-java-sdk-kms:1.12.538")
+
+  testImplementation(kotlin("test"))
 }
 
 tasks.test {
--- a/crypto/build.gradle.kts
+++ b/crypto/build.gradle.kts
@@ -16,7 +16,7 @@ dependencies {
   implementation("org.bouncycastle:bcpkix-jdk15on:1.70")
   implementation(project(":common"))
 
-  implementation("com.amazonaws:aws-java-sdk-kms:1.12.538")
+  api("com.amazonaws:aws-java-sdk-kms:1.12.538")
 
   testImplementation(kotlin("test"))
 }
--- a/crypto/src/main/kotlin/web5/sdk/crypto/AwsKeyManager.kt
+++ b/crypto/src/main/kotlin/web5/sdk/crypto/AwsKeyManager.kt
@@ -1,0 +1,163 @@
+package web5.sdk.crypto
+
+import com.amazonaws.services.kms.AWSKMSClientBuilder
+import com.amazonaws.services.kms.model.CreateAliasRequest
+import com.amazonaws.services.kms.model.CreateKeyRequest
+import com.amazonaws.services.kms.model.DescribeKeyRequest
+import com.amazonaws.services.kms.model.GetPublicKeyRequest
+import com.amazonaws.services.kms.model.KeySpec
+import com.amazonaws.services.kms.model.KeyUsageType
+import com.amazonaws.services.kms.model.MessageType
+import com.amazonaws.services.kms.model.SignRequest
+import com.amazonaws.services.kms.model.SigningAlgorithmSpec
+import com.nimbusds.jose.Algorithm
+import com.nimbusds.jose.JWSAlgorithm
+import com.nimbusds.jose.JWSHeader
+import com.nimbusds.jose.JWSObject
+import com.nimbusds.jose.Payload
+import com.nimbusds.jose.crypto.impl.ECDSA
+import com.nimbusds.jose.jwk.Curve
+import com.nimbusds.jose.jwk.ECKey
+import com.nimbusds.jose.jwk.JWK
+import com.nimbusds.jose.jwk.KeyUse
+import org.bouncycastle.asn1.x509.SubjectPublicKeyInfo
+import org.bouncycastle.crypto.ExtendedDigest
+import org.bouncycastle.crypto.digests.SHA256Digest
+import org.bouncycastle.crypto.digests.SHA384Digest
+import org.bouncycastle.crypto.digests.SHA512Digest
+import org.bouncycastle.openssl.jcajce.JcaPEMKeyConverter
+import web5.sdk.common.Base58Btc
+import java.nio.ByteBuffer
+import java.security.PublicKey
+import java.security.interfaces.ECPublicKey
+
+/**
+ * A [KeyManager] that delegates to AWS KMS for storage of keys and signing operations. Key aliases are
+ * deterministically generated from base58btc encoding of the public key
+ */
+public class AwsKeyManager : KeyManager {
+
+  private val kmsClient = AWSKMSClientBuilder.standard().build()
+
+  override fun generatePrivateKey(algorithm: Algorithm, curve: Curve?, options: KeyGenOptions?): String {
+    val keySpec = getDetails(algorithm).keySpec
+    val createKeyRequest = CreateKeyRequest()
+      .withKeySpec(keySpec)
+      .withKeyUsage(KeyUsageType.SIGN_VERIFY)
+    val createKeyResponse = kmsClient.createKey(createKeyRequest)
+    val keyId = createKeyResponse.keyMetadata.keyId
+
+    val getPublicKeyRequest = GetPublicKeyRequest()
+      .withKeyId(keyId)
+    val publicKeyResponse = kmsClient.getPublicKey(getPublicKeyRequest)
+
+    val alias = generateKeyAlias(publicKeyResponse.publicKey)
+    val createAliasRequest = CreateAliasRequest()
+      .withAliasName(alias)
+      .withTargetKeyId(keyId)
+    kmsClient.createAlias(createAliasRequest)
+
+    return alias
+  }
+
+  override fun getPublicKey(keyAlias: String): JWK {
+    val getPublicKeyRequest = GetPublicKeyRequest()
+      .withKeyId(keyAlias)
+    val publicKeyResponse = kmsClient.getPublicKey(getPublicKeyRequest)
+
+    val details = getDetails(publicKeyResponse.keySpec.enum())
+    val publicKey = convertToJavaPublicKey(publicKeyResponse.publicKey)
+
+    val jwkBuilder = when (publicKey) {
+      is ECPublicKey -> ECKey.Builder(details.curve, publicKey)
+      else -> throw IllegalArgumentException("Unknown key type $publicKey")
+    }
+    return jwkBuilder
+      .algorithm(details.algorithm)
+      .keyID(keyAlias)
+      .keyUse(KeyUse.SIGNATURE)
+      .build()
+  }
+
+  /**
+   * Parse the ASN.1 DER encoded public key that AWS KMS returns, and convert it to a standard PublicKey
+   */
+  private fun convertToJavaPublicKey(publicKeyDerBytes: ByteBuffer): PublicKey {
+    val publicKeyInfo = SubjectPublicKeyInfo.getInstance(publicKeyDerBytes.array())
+    return JcaPEMKeyConverter().getPublicKey(publicKeyInfo)
+  }
+
+  override fun sign(keyAlias: String, signingInput: ByteArray): ByteArray {
+    val keySpec = fetchKeySpec(keyAlias)
+    val details = getDetails(keySpec)
+    //Pre-hash the input because AWS limits the message to 4096 bytes
+    val hashedMessage = shaDigest(details, signingInput)
+
+    val signRequest = SignRequest()
+      .withKeyId(keyAlias)
+      .withMessageType(MessageType.DIGEST)
+      .withMessage(hashedMessage.asByteBuffer())
+      .withSigningAlgorithm(details.signingAlgorithm)
+    val signResponse = kmsClient.sign(signRequest)
+    val derSignatureBytes = signResponse.signature.array()
+    return transcodeDerSignatureToConcat(derSignatureBytes, details.algorithm)
+  }
+
+  /**
+   * Fetch the [KeySpec] using a [DescribeKeyRequest]
+   */
+  private fun fetchKeySpec(keyAlias: String): KeySpec {
+    val describeKeyRequest = DescribeKeyRequest().withKeyId(keyAlias)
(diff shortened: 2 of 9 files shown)


crypto/src/main/kotlin/web5/sdk/crypto/AwsKeyManager.kt
```diff
@@ -0,0 +1,200 @@
...
+    val keySpec: KeySpec,
+    val signingAlgorithm: SigningAlgorithmSpec,
+    val newDigest: () -> ExtendedDigest
+  )
+
+  private val algorithmDetails = mapOf(
+    JWSAlgorithm.ES256K to Details(JWSAlgorithm.ES256K, Curve.SECP256K1, KeySpec.ECC_SECG_P256K1, SigningAlgorithmSpec.ECDSA_SHA_256) { SHA256Digest() },
+    JWSAlgorithm.ES256 to Details(JWSAlgorithm.ES256, Curve.P_256, KeySpec.ECC_NIST_P256, SigningAlgorithmSpec.ECDSA_SHA_256) { SHA256Digest() },
+    JWSAlgorithm.ES384 to Details(JWSAlgorithm.ES384, Curve.P_384, KeySpec.ECC_NIST_P384, SigningAlgorithmSpec.ECDSA_SHA_384) { SHA384Digest() },
+    JWSAlgorithm.ES512 to Details(JWSAlgorithm.ES512, Curve.P_521, KeySpec.ECC_NIST_P521, SigningAlgorithmSpec.ECDSA_SHA_512) { SHA512Digest() }
```
mistermoe, line 61: we should comment these out until we have support for them in `Crypto`. Otherwise we won't be able to verify signatures produced using these algos

mistermoe: can we pull the hashing functions into a reusable util?

crypto/src/main/kotlin/web5/sdk/crypto/AwsKeyManager.kt
```diff
@@ -0,0 +1,200 @@
...
+ * Key aliases are generated from the key's JWK thumbprint, and stored in AWS KMS.
+ * e.g. alias/6uNnyj7xZUgtKTEOFV2mz0f7Hd3cxIH1o5VXsOo4u1M
+ *
+ * AWS supports a limited set ECDSA curves for signing:
+ * - [JWSAlgorithm.ES256K]
+ * - [JWSAlgorithm.ES256]
+ * - [JWSAlgorithm.ES384]
+ * - [JWSAlgorithm.ES512]
+ */
+public class AwsKeyManager : KeyManager {
```
mistermoe, line 45: note from @bradleydwyer: might be worth including ability to optionally provide AWS KMS Client configuration

crypto/src/main/kotlin/web5/sdk/crypto/AwsKeyManager.kt
```diff
@@ -0,0 +1,200 @@
...
+
+  private val kmsClient: AWSKMS = AWSKMSClientBuilder.standard().build()
```
mistermoe, line 47: note from @bradleydwyer: might be worth including ability to optionally provide AWS KMS Client configuration

crypto/src/test/kotlin/web5/sdk/crypto/AwsKeyManagerTest.kt
```diff
@@ -0,0 +1,70 @@
...
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.assertThrows
+import kotlin.test.assertEquals
+
+class AwsKeyManagerTest {
+
+  val signingInput = "The Magic Words are Squeamish Ossifrage".toByteArray()
+  val awsKeyManager = AwsKeyManager()
+
+//  @Test
```
phoebe-lew, line 20: intended to be commented out?

phoebe-lew (approved):

mistermoe (approved):
barring @phoebe-lew's comment, looks gud2me! 🏅

crypto/src/main/kotlin/web5/sdk/crypto/AwsKeyManager.kt
```diff
@@ -0,0 +1,223 @@
...
+ * [Configure the AWS CLI](https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-configure.html)
+ *
+ * Key aliases are generated from the key's JWK thumbprint, and stored in AWS KMS.
+ * e.g. alias/6uNnyj7xZUgtKTEOFV2mz0f7Hd3cxIH1o5VXsOo4u1M
+ *
+ * AWSKeyManager supports a limited set ECDSA curves for signing:
+ * - [JWSAlgorithm.ES256K]
+ */
+public class AwsKeyManager(
+  private val kmsClient: AWSKMS = AWSKMSClientBuilder.standard().build()
```
bradleydwyer, line 42: Consider allowing the kmsClient to be provided by the caller (e.g. to allow configurable behaviour of the client like timeouts)

crypto/src/main/kotlin/web5/sdk/crypto/AwsKeyManager.kt
```diff
@@ -0,0 +1,223 @@
...
+  private data class AlgorithmDetails(
+    val algorithm: JWSAlgorithm,
+    val curve: Curve,
+    val keySpec: KeySpec,
+    val signingAlgorithm: SigningAlgorithmSpec,
+    val newDigest: () -> ExtendedDigest
+  )
+
+  private val algorithmDetails = mapOf(
+    JWSAlgorithm.ES256K to AlgorithmDetails(
```
bradleydwyer, line 54: As a note, it is possible to drop the enum type for the assignment, which might be useful given the use of named params. e.g.

```    ES256K to AlgorithmDetails(
      algorithm = ES256K,
      curve = SECP256K1,
      keySpec = ECC_SECG_P256K1,
      signingAlgorithm = ECDSA_SHA_256,
      newDigest = { SHA256Digest() }
    )
```

bradleydwyer (approved):

crypto/src/test/kotlin/web5/sdk/crypto/AwsKeyManagerTest.kt
tomdaffurn: Needs AWS creds, so won't work in GHA. I tried a few different ways of disabling by annotation, and gave up in frustration.

mistermoe: is `@Ignore` or `@Disabled` not working because we haven't pulled junit in as a dependency?

tomdaffurn: Both annotations are working locally for me. Trying to match versions of Gradle and Java to repro the problem, but not having any luck

tomdaffurn: We have
```
tasks.test {
  useJUnitPlatform()
}
```

mistermoe: tested by adding `testImplementation("org.junit.jupiter:junit-jupiter:5.8.1")` as a dependency then tried using `@Disabled` and `@EnabledIfEnvironmentVariable(named = "AWS", matches = "something")`. both worked

image

crypto/build.gradle.kts
```diff
@@ -15,6 +15,9 @@ dependencies {
   implementation("org.bouncycastle:bcprov-jdk15on:1.70")
   implementation("org.bouncycastle:bcpkix-jdk15on:1.70")
   implementation(project(":common"))
+
+  implementation("com.amazonaws:aws-java-sdk-kms:1.12.538")
```
andresuribe87, line 19: It seems like AWS specific definitions (for example, `AWSKMS` as the constructor parameter) are needed for client to be able to call this. Thus, it's making part of the API contract, and you'd need to change this. 
```suggestion
  api("com.amazonaws:aws-java-sdk-kms:1.12.538")
```

crypto/src/main/kotlin/web5/sdk/crypto/AwsKeyManager.kt
```diff
@@ -0,0 +1,223 @@
...
+  }
+
+  /**
+   * Return the alias of [publicKey], as was originally returned by [generatePrivateKey].
+   *
+   * @param publicKey A public key in JWK (JSON Web Key) format
+   * @return The alias belonging to [publicKey]
+   */
+  override fun getDefaultAlias(publicKey: JWK): String {
+    val jwkThumbprint = publicKey.computeThumbprint()
```
andresuribe87, line 174: Curious why thumbprint instead of using the `kid` field?

crypto/src/main/kotlin/web5/sdk/crypto/KeyManager.kt
```diff
@@ -46,11 +45,21 @@ public interface KeyManager {
...
-   * @param payload The data/payload to be signed.
+   * @param signingInput The data to be signed.
+   * @return The signature in JWS R+S format
    *
    * Implementations should ensure that the signing process is secured, utilizing secure cryptographic
    * practices and safeguarding the private key during the operation. The specific signing algorithm
    * used may depend on the type and parameters of the stored key.
    */
-  public fun sign(keyAlias: String, payload: Payload)
+  public fun sign(keyAlias: String, signingInput: ByteArray): ByteArray
```
andresuribe87, line 55: I think this would make it easier to consume for everyone.
```suggestion
  public fun sign(keyAlias: String, signingInput: ByteArray): JWSObject
```

crypto/src/main/kotlin/web5/sdk/crypto/KeyManager.kt
```diff
@@ -46,11 +45,21 @@ public interface KeyManager {
...
+
+  /**
+   * Return the alias of [publicKey], as was originally returned by [generatePrivateKey].
+   *
+   * @param publicKey A public key in JWK (JSON Web Key) format
+   * @return The alias belonging to [publicKey]
+   * @throws IllegalArgumentException if the key is not known to the [KeyManager]
+   */
+  public fun getDefaultAlias(publicKey: JWK): String
```
andresuribe87, line 64: Given that `JWK` already have a `kid` field, is it possible to use that field to identify keys?

IMO, it's easier to put the alias within that field, rather than expanding the interface.

crypto/src/test/kotlin/web5/sdk/crypto/AwsKeyManagerTest.kt
```diff
@@ -0,0 +1,70 @@
...
+import com.nimbusds.jose.jwk.KeyUse
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.assertThrows
+import kotlin.test.Ignore
+import kotlin.test.assertEquals
+
+class AwsKeyManagerTest {
+
+  val signingInput = "The Magic Words are Squeamish Ossifrage".toByteArray()
+  val awsKeyManager = AwsKeyManager()
```
andresuribe87, line 18: Does this mean that this unit test is making network calls? Curious how this works.

andresuribe87 (approved):

crypto/src/test/kotlin/web5/sdk/crypto/AwsKeyManagerTest.kt
tomdaffurn: It doesn't 😃 Fighting with gradle/junit to disable them in GHA now

crypto/src/main/kotlin/web5/sdk/crypto/AwsKeyManager.kt
tomdaffurn: It's already possibly for the caller to supply, isn't it?
```
  fun `test a custom KMS client`() {
    val kmsClient = AWSKMSClient.builder()
      .withCredentials(AWSStaticCredentialsProvider(BasicAWSCredentials("foo", "bar")))
      .build()
    val customisedKeyManager = AwsKeyManager(kmsClient = kmsClient)

    assertThrows<AmazonServiceException> {
      customisedKeyManager.generatePrivateKey(JWSAlgorithm.ES256K)
    }
  }
```

crypto/src/main/kotlin/web5/sdk/crypto/KeyManager.kt
tomdaffurn: There was an argument made that all crypto ops should be bytes-in/bytes-out. This makes KeyManager compatible with Crypto.kt

crypto/src/test/kotlin/web5/sdk/crypto/AwsKeyManagerTest.kt
tomdaffurn: Resolved it. I made a silly mistake by instantiating AwsKeyManager outside the test methods

crypto/src/main/kotlin/web5/sdk/crypto/KeyManager.kt
tomdaffurn: Yeah, this a good point, and kicked off a big discussion. Both AwsKeyManager and InMemoryKeyManager are setting the alias into `kid`, so it would work with the code we have here.
 
It's not clear that the publicKeyJwk will always have a `kid` though. It's optional. I think it's up to the DID method how that publicKeyJwk is put together?

tomdaffurn: I'm gonna merge this now to unblock #63, but happy to keep the convo going in Slack

crypto/src/main/kotlin/web5/sdk/crypto/AwsKeyManager.kt
tomdaffurn: This does go into `kid` up in `generatePrivateKey`

mistermoe: @andresuribe87 because `kid` isn't a required property of JWKs ([reference](https://datatracker.ietf.org/doc/html/rfc7517#section-4.5)). 

IMO, Using [JWK thumbprints](https://datatracker.ietf.org/doc/html/rfc7638) makes it so there's a deterministic mapping between a DID's verification method (presuming that the verification method has `publicKeyJwk`, a requirement we intend to surface as part of an upcoming interop profile) and a key manager's key alias without having to rely on a [Verification Method ID](https://www.w3.org/TR/did-core/#verification-methods) or [JWK kid](https://datatracker.ietf.org/doc/html/rfc7517#section-4.5). Especially because a verification method id and a kid (even if both are present) don't have to match

Issue #48, opened 2023-10-03
[Bug] web5-sdk-kotlin forces use of Kotlin 1.9.0


tomdaffurn:
**Describe the bug**
When using web5-sdk-kotlin in a project using Kotlin version 1.7.20, the build fails:
> Class 'kotlin.Unit' was compiled with an incompatible version of Kotlin. The binary version of its metadata is 1.9.0, expected version is 1.7.1

Upgrading to Kotlin 1.9.0 resolves the issue.

**To Reproduce:**

1. Clone tbdiligence:
```
git clone git@github.com:tbdeng/tbdiligence.git
```

2. Add dependency on web5-sdk-kotlin:
```
repositories {
  maven("https://jitpack.io")
  maven("https://repo.danubetech.com/repository/maven-public/")
}
dependencies {
  implementation("com.github.TBD54566975.web5-sdk-kotlin:dids:v0.0.0-alpha")
}
```

3. Build it
```
./gradlew clean build
```

**Expected behavior:**
Successful build

**Supporting Material**
```
> Task :api:compileKotlin FAILED
'compileJava' task (current target is 11) and 'compileKotlin' task (current target is 1.8) jvm target compatibility should be set to the same Java version.
e: Incompatible classes were found in dependencies. Remove them from the classpath or use '-Xskip-metadata-version-check' to suppress errors
e: /Users/tdaffurn/.gradle/caches/modules-2/files-2.1/com.github.TBD54566975.web5-sdk-kotlin/dids/v0.0.0-alpha/9d974df986b1cc0ea21cb93e609607e7a1853136/dids-v0.0.0-alpha.jar!/META-INF/dids.kotlin_module: Module was compiled with an incompatible version of Kotlin. The binary version of its metadata is 1.9.0, expected version is 1.7.1.
```

**Environment (please complete the following information):**

MacOS Sonoma
Kotlin 1.7.20
openjdk version "11.0.16.1" 2022-08-12

nitro-neal:
@mistermoe we good on this one?

PR #76, opened 2023-10-10
downgrading kotlin

nitro-neal:
# Overview
Addresses https://github.com/TBD54566975/web5-kt/issues/48

# Description
TBDiligence uses:
image


andresuribe87 (approved):

mistermoe:
Yoo! i might be missing context here bc i'm not sure what y'all tried before this, i know y'all tried quite a bit of stuff. I feel like this fixes the problem for tbdiligence but it seems strange that the minor version of kotlin we're using for this lib must match the minor version of kotlin that downstream consumers are using.

feels like there oughta be some compatibility option like "compile such that it works with minimum 1.7.0" or something along those lines.

otherwise we're going to end up in a situation where we're constantly changing versions depending on the kotlin version that  our downstream consumers are using

mistermoe:
@nitro-neal looks like we can specify kotlin version compat - [docs here](https://kotlinlang.org/docs/gradle-compiler-options.html#attributes-common-to-jvm-and-js)

image


in the root `build.gradle.kts` within `subprojects` might be worth trying:

```
  kotlin {
    explicitApi()
    jvmToolchain(11)
    compilerOptions {
      jvmTarget.set(JvmTarget.JVM_11)
      apiVersion.set(org.jetbrains.kotlin.gradle.dsl.KotlinVersion.KOTLIN_1_7)
      languageVersion.set(org.jetbrains.kotlin.gradle.dsl.KotlinVersion.KOTLIN_1_7)
    }
  }

  java {
    targetCompatibility = JavaVersion.VERSION_11
    withJavadocJar()
    withSourcesJar()
  }
```

nitro-neal:
no good, when I try to do the above and consume it with our test app
still getting:

Module was compiled with an incompatible version of Kotlin. The binary version of its metadata is 1.9.0, expected version is 1.7.1.

nitro-neal:
Also I need to confirm, but I think that if any consuming app is our version or later it will work, doesn't have to be the exact same version

mistermoe:
yoo @nitro-neal i think we can close this out yeah?

PR #77, opened 2023-10-11
DID ION Update, Recover, and Deactivate Operations

### LINKED ISSUE #1: Implement did:ion Update Operation
**Background/context**: [click here](https://github.com/TBD54566975/web5-js/issues/176).

### LINKED ISSUE #3: Implement did:ion Deactivate Operation
**Background/context**: [click here](https://github.com/TBD54566975/web5-js/issues/181).

andresuribe87:
# Overview
Fixes #1 by implementing did ion updates.

Includes the updates of PR #88, which fixes #3 and #2. 

# Description

After this PR, all ion type of updates will be supported.

# How Has This Been Tested?
Unit test that mimics https://github.com/decentralized-identity/ion-sdk/blob/main/tests/IonRequest.spec.ts#L38 as much as possible.


crypto/src/main/kotlin/web5/sdk/crypto/Crypto.kt
```diff
@@ -106,7 +103,11 @@ public object Crypto {
     val rawCurve = privateKey.toJSONObject()["crv"]
     val curve = rawCurve?.let { Curve.parse(it.toString()) }
 
-    val signer = getSigner(privateKey.algorithm, curve)
+    val signer = if (curve == Curve.SECP256K1 && privateKey.algorithm == null) {
```
nitro-neal, line 106: dont understand this, maybe a comment on why if the alg is null we hardcode it

crypto/src/main/kotlin/web5/sdk/crypto/KeyManager.kt
```diff
@@ -52,7 +53,7 @@ public interface KeyManager {
    * practices and safeguarding the private key during the operation. The specific signing algorithm
    * used may depend on the type and parameters of the stored key.
    */
-  public fun sign(keyAlias: String, payload: Payload)
+  public fun sign(keyAlias: String, payload: Payload): JWSObject
```
nitro-neal, line 56: I think you may have some merge conflicts with the pr because I see:

https://github.com/TBD54566975/web5-kt/pull/63/files#diff-a76ca21a2d268f0c593a56f6a40fa7ab54cabe5405bbc08a37e14ffee6e74006R93

    val signatureBytes = did.keyManager.sign(keyAlias, toSign)

dids/src/main/kotlin/web5/sdk/dids/DIDIonManager.kt
```diff
@@ -168,6 +214,64 @@ public sealed class DIDIonManager(
...
+      }
+    }
+    val opBody = runBlocking { response.bodyAsText() }
+    if (response.status.isSuccess()) {
+      return IonUpdateMetadata(
+        operationsResponseBody = opBody,
+        updateKeyAlias = newUpdateKeyAlias,
+      )
+    }
+    throw Exception("received error response '$opBody'")
```
nitro-neal, line 235:
```suggestion
    throw Exception("received error response: '$opBody'")
```

nitro-neal (approved):
Looks good, just make sure the sign is working the same way as https://github.com/TBD54566975/web5-kt/pull/63

crypto/src/main/kotlin/web5/sdk/crypto/Crypto.kt
andresuribe87: Reverted this changes in favor of adding them in the mapping.

crypto/src/main/kotlin/web5/sdk/crypto/KeyManager.kt
andresuribe87: Yes, they've been fixed now.

andresuribe87 (commented):
With the latest changes from the base branch now

andresuribe87:
@bradleydwyer you made comments on a very old commit. Can you take another stab at it, please?

dids/src/main/kotlin/web5/sdk/dids/DidIonManager.kt
```diff
@@ -62,6 +75,42 @@ public fun DidIonManager(builderAction: DidIonConfiguration.() -> Unit): DidIonM
...
+ * @param updateKeyAlias The alias within the key manager that refers to the last update key.
+ * @param servicesToAdd The services to add in the did document.
+ * @param idsOfServicesToRemove Ids of the services to remove from the did document.
+ * @param publicKeysToAdd PublicKeys to add to the DID document.
+ * @param idsOfPublicKeysToRemove Keys to remove from the DID document.
+ */
+public class UpdateDidIonOptions(
+  public val didString: String,
+  public val updateKeyAlias: String,
+  public val servicesToAdd: Iterable<Service>? = null,
```
bradleydwyer, line 91: I think it might be preferable to just create empty collections by default rather than cater for the possibility of a null collection (https://github.com/TBD54566975/web5-kt/pull/77#discussion_r1354622606)

dids/src/main/kotlin/web5/sdk/dids/DidIonManager.kt
```diff
@@ -62,6 +75,42 @@ public fun DidIonManager(builderAction: DidIonConfiguration.() -> Unit): DidIonM
...
+ * The options when updating an ION did.
+ *
+ * @param didString The full did. e.g. `did:ion:123`
+ * @param updateKeyAlias The alias within the key manager that refers to the last update key.
+ * @param servicesToAdd The services to add in the did document.
+ * @param idsOfServicesToRemove Ids of the services to remove from the did document.
+ * @param publicKeysToAdd PublicKeys to add to the DID document.
+ * @param idsOfPublicKeysToRemove Keys to remove from the DID document.
+ */
+public class UpdateDidIonOptions(
```
bradleydwyer, line 88: A possible alternative to this impl might be

```
public data class UpdateDidIonOptions(
  val didString: String,
  val updateKeyAlias: String,
  val servicesToAdd: List<Service> = emptyList(),
  val idsOfServicesToRemove: List<String> = emptyList(),
  val publicKeysToAdd: List<PublicKey> = emptyList(),
  val idsOfPublicKeysToRemove: List<String> = emptyList()
) {
  internal fun toPatches(): List<PatchAction> {
    fun <T> MutableList<PatchAction>.addIfNotEmpty(list: List<T>, action: (List<T>) -> PatchAction) {
      list.takeIf { it.isNotEmpty() }?.let { this.add(action(it)) }
    }

    return buildList {
      addIfNotEmpty(servicesToAdd, ::AddServicesAction)
      addIfNotEmpty(idsOfServicesToRemove, ::RemoveServicesAction)
      addIfNotEmpty(publicKeysToAdd, ::AddPublicKeysAction)
      addIfNotEmpty(idsOfPublicKeysToRemove, ::RemovePublicKeysAction)
    }
  }
}
```
For a rough and untested stab at it. I think this covers the comments below.

(https://github.com/TBD54566975/web5-kt/pull/77#discussion_r1354623761)

dids/src/main/kotlin/web5/sdk/dids/DidIonManager.kt
bradleydwyer, line 91: Curious as to why the adds are more generic than the remove (Iterable vs Set). Is order important for add operations?

(https://github.com/TBD54566975/web5-kt/pull/77#discussion_r1354395800)

dids/src/main/kotlin/web5/sdk/dids/ion/model/Models.kt
```diff
@@ -151,6 +169,11 @@ public data class OperationSuffixDataObject(
  */
 public typealias Commitment = String
 
+/**
+ * Type alias for reveal value.
+ */
+public typealias Reveal = String
```
bradleydwyer, line 175: Nit: IMHO these comments are super obvious. I don't think they add value here and make the code less readable.

(https://github.com/TBD54566975/web5-kt/pull/77#discussion_r1354478198)

dids/src/main/kotlin/web5/sdk/dids/ion/model/Models.kt
```diff
@@ -102,7 +102,10 @@ public enum class PublicKeyPurpose(@get:JsonValue public val code: String) {
 )
 @JsonSubTypes(
   JsonSubTypes.Type(AddServicesAction::class, name = "add-services"),
-  JsonSubTypes.Type(ReplaceAction::class, name = "replace")
+  JsonSubTypes.Type(ReplaceAction::class, name = "replace"),
+  JsonSubTypes.Type(RemoveServicesAction::class, name = "remove-services"),
+  JsonSubTypes.Type(AddPublicKeysAction::class, name = "add-public-keys"),
+  JsonSubTypes.Type(RemovePublicKeysAction::class, name = "remove-public-keys"),
 )
 public sealed class PatchAction
```
bradleydwyer, line 110: Curious why the sealed class route vs a marker interface?

(https://github.com/TBD54566975/web5-kt/pull/77#discussion_r1354476007)

dids/src/main/kotlin/web5/sdk/dids/DidIonManager.kt
bradleydwyer: When I looked at the spec it didn't seem like the order was important (or that each type of patch was required to be present)

bradleydwyer:
> @bradleydwyer you made comments on a very old commit. Can you take another stab at it, please?

They were made 5 days ago, but held off hitting submit on the review to avoid distracting from deadline. Have duplicated the comment on the current HEAD with pointers back.

dids/src/main/kotlin/web5/sdk/dids/ion/model/Models.kt
```diff
@@ -151,6 +169,11 @@ public data class OperationSuffixDataObject(
  */
 public typealias Commitment = String
```
bradleydwyer, line 170: Declared but not used in places like

```
public class MetadataMethod(
  public val published: Boolean,
  public val recoveryCommitment: String,
  public val updateCommitment: String,
)
```

which I presume it should be?

dids/src/main/kotlin/web5/sdk/dids/ion/model/Models.kt
```diff
@@ -161,6 +184,27 @@ public data class SidetreeCreateOperation(
 
 }
 
+/**
+ * Sidetree update operation as defined in https://identity.foundation/sidetree/api/#update
+ */
+public data class SidetreeUpdateOperation(
+  public val type: String,
+  public val didSuffix: String,
+  public val revealValue: String,
```
bradleydwyer, line 193: use the `Reveal` typealias?

dids/src/test/kotlin/web5/sdk/dids/DidIonTest.kt
```diff
@@ -124,6 +132,76 @@ class DidIonTest {
     }
   }
 
+  @Test
+  fun `update fails when update key is absent`() {
+    val result = assertThrows<Exception> {
```
bradleydwyer, line 137: `InvalidStatusException` instead?

dids/src/main/kotlin/web5/sdk/dids/DidIonManager.kt
```diff
@@ -287,12 +406,14 @@ public sealed class DidIonManager(
     val canonicalized = JsonCanonicalizer(pkJson).encodedUTF8
 
     // 3. Use the implementation’s HASH_PROTOCOL to Multihash the canonicalized public key to generate the REVEAL_VALUE,
-    val intermediate = Multihash.sum(Multicodec.SHA2_256, canonicalized).getOrThrow().digest
+    val mh = Multihash.sum(Multicodec.SHA2_256, canonicalized).getOrThrow()
+    val reveal = Convert(mh.bytes()).toBase64Url(padding = false)
```
bradleydwyer, line 410: Suggest explicitly adding `Reveal` as a type.

dids/src/main/kotlin/web5/sdk/dids/DidIonManager.kt
bradleydwyer: And in addition to https://github.com/TBD54566975/web5-kt/pull/77/files#r1359995250 was confused at marshalling everything into a List at the end - why not just be a List from the start?

dids/src/main/kotlin/web5/sdk/dids/ion/model/Models.kt
bradleydwyer, line 1: For the jackson deserializer

```
object Deserializer : JsonDeserializer<JWK>() {
    override fun deserialize(p: JsonParser, ctxt: DeserializationContext?): JWK {
      @Suppress("UNCHECKED_CAST")
      val node = p.readValueAs(Map::class.java) as MutableMap<String, Any>
      return JWK.parse(node)
    }
  }
```
You can probably avoid the class cast suppression and also avoid type erasure issues with the `readValueAs` call with the following:

```
  object Deserializer : JsonDeserializer<JWK>() {
    override fun deserialize(p: JsonParser, ctxt: DeserializationContext?): JWK {
      val typeRef = object : TypeReference<HashMap<String, Any>>() {}
      val node = p.readValueAs(typeRef) as HashMap<String, Any>
      return JWK.parse(node)
    }
  }
```

phoebe-lew (approved):

dids/src/main/kotlin/web5/sdk/dids/DIDIonManager.kt
```diff
@@ -61,6 +71,42 @@ public fun DIDIonManager(builderAction: DIDIonConfiguration.() -> Unit): DIDIonM
...
+ * @param publicKeysToAdd PublicKeys to add to the DID document.
+ * @param idsOfPublicKeysToRemove Keys to remove from the DID document.
+ */
+public class UpdateDidIonOptions(
+  public val didString: String,
+  public val updateKeyAlias: String,
+  public val servicesToAdd: Iterable<Service>? = null,
+  public val idsOfServicesToRemove: Set<String>? = null,
+  public val publicKeysToAdd: Iterable<PublicKey>? = null,
+  public val idsOfPublicKeysToRemove: Set<String>? = null,
```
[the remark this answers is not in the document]
mistermoe: agreed

crypto/src/main/kotlin/web5/sdk/crypto/InMemoryKeyManager.kt
```diff
@@ -90,9 +90,6 @@ public class InMemoryKeyManager : KeyManager {
    * Imports [jwk] and returns the alias that refers to it.
    */
   public fun import(jwk: JWK): String {
-    require(jwk.isPrivate) {
-      "Importing a non-private key is not permitted"
-    }
```
mistermoe, line 95: what's the use-case driving the ability to import a public key?

mistermoe:
Took some time to use the current ION implementation last night beyond the default scenario. 

There's quite a bit of optionality that ends up creating a strange usage pattern where the caller effectively ends up having to use the keyManger _outside_ ion to create keys _for or associated to_  an ion did. I think it'd make more sense for the optionality to look something like this:

```kt
package web5.sdk.dids
import web5.sdk.crypto.InMemoryKeyManager

public fun main() {
  val km = InMemoryKeyManager()
  val opts = CreateDidIonOptions(
    services = listOf(
      Service(...)
    ),
    verificationMethods = listOf(
      IonVerificationMethod(
        alg: JWSAlgorithm.ES256K,
        relationships = listOf(VerificationRelationships.Authentication, VerificationRelationships.Assertion)
      )
    )
  )

  val did = DidIon.create(km)
}
```

other thoughts:
* found the name `DidIonManager` to be confusing. instinctively assuming that it _managed_ multiple ion dids. or was persistently stateful in some way. personally prefer `DidIon`
* `DidIonHandle` was also a bit confusing. general pattern that was put in place was any concrete. the general pattern that seemed to stick was the way in which `DidKey` is implemented
* would be nice to use `okhttp` instead of `ktor` so that there's some congruency across kotlin repos. tbh though i haven't dug into kotlin multiplatform enough to know whether ktor actually helps here or not in a way that okhttp doesn't.


I think we can merge this PR into main and then address the above as a whole in a separate PR. thoughts?

dids/src/main/kotlin/web5/sdk/dids/ion/model/Models.kt
```diff
@@ -163,6 +233,38 @@ public data class SidetreeCreateOperation(
...
+  public val delta: Delta,
+  public val signedData: String,
+)
+
+/**
+ * Update operation signed data object as defined in https://identity.foundation/sidetree/spec/#update-signed-data-object
+ */
+public data class UpdateOperationSignedData(
+  @JsonSerialize(using = JacksonJwk.Serializer::class)
+  @JsonDeserialize(using = JacksonJwk.Deserializer::class)
```
tomdaffurn, line 263: You can create a Jackson module that is registered on the ObjectMapper once, instead of annotating every use of a type
https://github.com/TBD54566975/tbdex-kt/blob/main/protocol/src/main/kotlin/tbdex/sdk/protocol/serialization/Json.kt#L35

credentials/src/main/kotlin/web5/sdk/credentials/VerifiableCredential.kt
```diff
@@ -72,7 +72,12 @@ public class VerifiableCredential(public val vcDataModel: VcDataModel) {
    */
   public fun sign(did: Did, assertionMethodId: String? = null): String {
     val didResolutionResult = DidResolvers.resolve(did.uri)
-    val assertionMethods = didResolutionResult.didDocument.assertionMethodVerificationMethodsDereferenced
+    val assertionMethods: List<VerificationMethod>? =
+      didResolutionResult.didDocument.assertionMethodVerificationMethodsDereferenced
+
+    require(!assertionMethods.isNullOrEmpty()) {
+      throw SignatureException("no assertion methods found in did document")
```
tomdaffurn, line 79:
```suggestion
      throw SignatureException("No assertion methods found in DID document")
```

tomdaffurn, line 79:
```suggestion
      throw SignatureException("No assertion methods found in DID document")
```

tomdaffurn (approved):

crypto/src/main/kotlin/web5/sdk/crypto/InMemoryKeyManager.kt
andresuribe87: I wanted to use  the same test vectors from the `ion-sdk` repository. The update request tests uses the public key of the next update, but not the private key. 

See https://github.com/TBD54566975/web5-kt/blob/ad6a91e1a8cac29c752156116fdf253395c5bf7c/dids/src/test/kotlin/web5/sdk/dids/DidIonTest.kt#L292 for where it's used. 

And see https://github.com/decentralized-identity/ion-sdk/blob/main/tests/IonRequest.spec.ts#L48 for the original test.

dids/src/main/kotlin/web5/sdk/dids/ion/model/Models.kt
andresuribe87: Thanks! I'm torn. On one hand, it's nice to not have to annotate all usages, but on the other hand, allowing developers to use this class with their own mapper without the need to register a custom module is also nice.

I'd like to keep as is, just because the devex experience is nicer. Happy to revisit!

andresuribe87:
thanks for the feedback @mistermoe 
Will merge and we can address in a separate PR. Some thoughts for now below

> Took some time to use the current ION implementation last night beyond the default scenario.
> 
> There's quite a bit of optionality that ends up creating a strange usage pattern where the caller effectively ends up having to use the keyManger _outside_ ion to create keys _for or associated to_ an ion did. I think it'd make more sense for the optionality to look something like this:
> 
> ```kotlin
> package web5.sdk.dids
> import web5.sdk.crypto.InMemoryKeyManager
> 
> public fun main() {
>   val km = InMemoryKeyManager()
>   val opts = CreateDidIonOptions(
>     services = listOf(
>       Service(...)
>     ),
>     verificationMethods = listOf(
>       IonVerificationMethod(
>         alg: JWSAlgorithm.ES256K,
>         relationships = listOf(VerificationRelationships.Authentication, VerificationRelationships.Assertion)
>       )
>     )
>   )
> 
>   val did = DidIon.create(km)
> }
> ```

I agree that there is too much optionality. That said, I'm not a fan of digressing from the [VerificationMethod data model](https://www.w3.org/TR/did-core/#verification-methods) described in the spec. Let's chat more about it in the followup PR.

> 
> other thoughts:
> 
> * found the name `DidIonManager` to be confusing. instinctively assuming that it _managed_ multiple ion dids. or was persistently stateful in some way. personally prefer `DidIon`

Curious what made you think there wasn't any state within the manager? 
- It is stateful (the httpclient and the ionHost are part of the state that's kept).
- It also has the ability to manage multiple dids (you can update any did).
 
My take on `DidIon` is that it doesn't tell me much about which actions I might call from the object. It sounds like you're proposing to use it like a namespace, rather than an actual object?

I originally thought to use `DidIonFactory`, but decided against it as a manager can update\recover\deactivate. See additional comments into the next point.

> * `DidIonHandle` was also a bit confusing. the general pattern that seemed to stick was the way in which `DidKey` is implemented

I actually found `DidKey` pattern to be confusing :P ; my mental model is that a `Did` object should follow what's defined as a DID in the [spec](https://www.w3.org/TR/did-core/#dfn-decentralized-identifiers)). As such, a DID isn't aware of it's context, nor does it know what private keys it has access to. It's simply an identifier.

> * would be nice to use `okhttp` instead of `ktor` so that there's some congruency across kotlin repos. tbh though i haven't dug into kotlin multiplatform enough to know whether ktor actually helps here or not in a way that okhttp doesn't.

No preference whatsoever. I did an example PR in https://github.com/TBD54566975/web5-kt/pull/86 showcasing what it would look like to support any http framework, which I discourage as it seems like overkill.

> 
> I think we can merge this PR into main and then address the above as a whole in a separate PR. thoughts?

PR #84, opened 2023-10-13
adding cred status list

nitro-neal:
# Overview
Cred Credential Status Lists for Revocation (or Suspension)

# Description
This change allows for generation of a status list credential. *NOTE* This is a stateless SDK and it is up to the issuer to set the IDs and host these creds for public retrieval

```
  /**
   * Create a [StatusListCredential] with a specific purpose, e.g., for revocation.
   *
   * @param statusListCredentialId The id used for the resolvable path to the status list credential [String].
   * @param issuer The issuer URI of the credential, as a [String].
   * @param statusPurpose The status purpose of the status list cred, eg: revocation, as a [StatusPurpose].
   * @param issuedCredentials The credentials to be included in the status list credential, eg: revoked credentials, list of type [VerifiableCredential].
   * @return A [VerifiableCredential] instance.
   *
   * Example:
   * ```
   * val statusListCredential = StatusListCredential.create("http://example.com/statuslistcred/id123", "http://example.com/issuers/1", StatusPurpose.REVOCATION, listOf(vc1,vc2))
   * ```
   */
```

## References
https://www.w3.org/TR/vc-status-list/

### DIFF
--- a/credentials/src/main/kotlin/web5/sdk/credentials/StatusListCredential.kt
+++ b/credentials/src/main/kotlin/web5/sdk/credentials/StatusListCredential.kt
@@ -1,0 +1,301 @@
+package web5.sdk.credentials
+
+import com.danubetech.verifiablecredentials.CredentialSubject
+import io.ktor.client.HttpClient
+import io.ktor.client.engine.cio.CIO
+import io.ktor.client.plugins.ClientRequestException
+import io.ktor.client.plugins.ResponseException
+import io.ktor.client.request.get
+import io.ktor.client.statement.bodyAsText
+import io.ktor.http.isSuccess
+import java.io.ByteArrayInputStream
+import java.io.ByteArrayOutputStream
+import java.net.URI
+import java.util.Base64
+import java.util.BitSet
+import java.util.Date
+import java.util.zip.GZIPInputStream
+import java.util.zip.GZIPOutputStream
+
+/**
+ * Type alias representing the danubetech Status List 2021 Entry data model.
+ * This typealias simplifies the use of the [com.danubetech.verifiablecredentials.credentialstatus.StatusList2021Entry] class.
+ */
+public typealias StatusList2021Entry = com.danubetech.verifiablecredentials.credentialstatus.StatusList2021Entry
+
+/**
+ * Status purpose of a status list credential or a credential with a credential status.
+ */
+public enum class StatusPurpose {
+  REVOCATION,
+  SUSPENSION
+}
+
+/**
+ * The JSON property key for an encoded list.
+ */
+public const val ENCODED_LIST: String = "encodedList"
+
+/**
+ * The JSON property key for a status purpose.
+ */
+public const val STATUS_PURPOSE: String = "statusPurpose"
+
+/**
+ * `StatusListCredential` represents a digitally verifiable status list credential according to the
+ * [W3C Verifiable Credentials Status List v2021](https://www.w3.org/TR/vc-status-list/).
+ *
+ * When a status list is published, the result is a verifiable credential that encapsulates the status list.
+ *
+ */
+public object StatusListCredential {
+  /**
+   * Create a [StatusListCredential] with a specific purpose, e.g., for revocation.
+   *
+   * @param statusListCredentialId The id used for the resolvable path to the status list credential [String].
+   * @param issuer The issuer URI of the credential, as a [String].
+   * @param statusPurpose The status purpose of the status list cred, eg: revocation, as a [StatusPurpose].
+   * @param issuedCredentials The credentials to be included in the status list credential, eg: revoked credentials, list of type [VerifiableCredential].
+   * @return A [VerifiableCredential] instance.
+   *
+   * Example:
+   * ```
+   * val statusListCredential = StatusListCredential.create("http://example.com/statuslistcred/id123", "http://example.com/issuers/1", StatusPurpose.REVOCATION, listOf(vc1,vc2))
+   * ```
+   */
+  public fun create(
+    statusListCredentialId: String,
+    issuer: String,
+    statusPurpose: StatusPurpose,
+    issuedCredentials: List<VerifiableCredential>
+  ): VerifiableCredential {
+    val statusListIndexes = prepareCredentialsForStatusList(statusPurpose, issuedCredentials)
+    val bitString = bitstringGeneration(statusListIndexes)
+
+    val claims = mapOf(STATUS_PURPOSE to statusPurpose.toString().lowercase(), ENCODED_LIST to bitString)
+    val credSubject = CredentialSubject.builder()
+      .id(URI.create(statusListCredentialId))
+      .type("StatusList2021")
+      .claims(claims)
+      .build()
+
+    val vcDataModel = VcDataModel.builder()
+      .id(URI.create(statusListCredentialId))
+      .issuer(URI.create(issuer))
+      .issuanceDate(Date())
+      .context(URI.create("https://w3id.org/vc/status-list/2021/v1"))
+      .type("StatusList2021Credential")
+      .credentialSubject(credSubject)
+      .build()
+
+    return VerifiableCredential(vcDataModel)
+  }
+
+  /**
+   * Validates if a given credential is part of the status list represented by a [VerifiableCredential].
+   *
+   * @param credentialToValidate The [VerifiableCredential] to be validated against the status list.
+   * @param statusListCredential The [VerifiableCredential] representing the status list.
+   * @return A [Boolean] indicating whether the `credentialToValidate` is part of the status list.
+   *
+   * This function checks if the given `credentialToValidate`'s status list index is present in the expanded status list derived from the `statusListCredential`.
+   *
+   * Example:
+   * ```
+   * val isRevoked = validateCredentialInStatusList(credentialToCheck, statusListCred)
+   * ```
+   */
+  public fun validateCredentialInStatusList(
+    credentialToValidate: VerifiableCredential,
+    statusListCredential: VerifiableCredential
+  ): Boolean {
+    val statusListEntryValue: StatusList2021Entry =
+      StatusList2021Entry.fromJsonObject(credentialToValidate.vcDataModel.credentialStatus.jsonObject)
+
+    val statusLisCredStatusPurpose: String? =
+      statusListCredential.vcDataModel.credentialSubject.jsonObject[STATUS_PURPOSE] as? String?
+
+    requireNotNull(statusListEntryValue.statusPurpose)
+    requireNotNull(statusLisCredStatusPurpose)
+    require(statusListEntryValue.statusPurpose == statusLisCredStatusPurpose)
+
+    val compressedBitstring: String? =
+      statusListCredential.vcDataModel.credentialSubject.jsonObject[ENCODED_LIST] as? String?
+
+    requireNotNull(compressedBitstring)
+    require(compressedBitstring.isNotEmpty())
+
+    val credentialIndex = statusListEntryValue.statusListIndex
+    val expandedValues: List<String> = bitstringExpansion(compressedBitstring)
+
+    return expandedValues.any { it == credentialIndex }
+  }
+
+  /**
+   * Validates if a given credential is part of the status list.
+   *
+   * @param credentialToValidate The [VerifiableCredential] to be validated against the status list.
+   * @param httpClient An optional [HttpClient] for fetching the status list credential. If not provided, a default HTTP client will be used.
+   * @return A [Boolean] indicating whether the `credentialToValidate` is part of the status list.
+   *
+   * This function fetches the status list credential from a URL present in the `credentialToValidate`.
+   * It supports using either a user-provided `httpClient` or a default client when no client is passed in.
+   * The function then checks if the given `credentialToValidate`'s status list index is present in the expanded status list derived from the fetched status list credential.
+   *
+   * Example:
+   * ```
+   * val isRevoked = validateCredentialInStatusList(credentialToCheck)
+   * ```
+   */
+  public suspend fun validateCredentialInStatusList(
+    credentialToValidate: VerifiableCredential,
+    httpClient: HttpClient? = null // default HTTP client but can be overridden
+  ): Boolean {
+
+    var isDefaultClient = false
+    val clientToUse = httpClient ?: defaultHttpClient().also { isDefaultClient = true }
+
+    try {
+      val statusListEntryValue: StatusList2021Entry =
+        StatusList2021Entry.fromJsonObject(credentialToValidate.vcDataModel.credentialStatus.jsonObject)
+      val statusListCredential =
+        clientToUse.fetchStatusListCredential(statusListEntryValue.statusListCredential.toString())
+
+      return validateCredentialInStatusList(credentialToValidate, statusListCredential)
+    } finally {
+      if (isDefaultClient) {
+        clientToUse.close()
+      }
+    }
+  }
+
+  private fun defaultHttpClient(): HttpClient {
+    return HttpClient(CIO)
+  }
+
+  private suspend fun HttpClient.fetchStatusListCredential(url: String): VerifiableCredential {
+    try {
+      val response: io.ktor.client.statement.HttpResponse = this.get(url)
+      if (response.status.isSuccess()) {
+        val body = response.bodyAsText()
+        return VerifiableCredential.parseJwt(body)
+      } else {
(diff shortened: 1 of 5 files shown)


credentials/src/main/kotlin/web5/sdk/credentials/StatusListCredential.kt
```diff
@@ -0,0 +1,301 @@
...
+    credentialToValidate: VerifiableCredential,
+    statusListCredential: VerifiableCredential
+  ): Boolean {
+    val statusListEntryValue: StatusList2021Entry =
+      StatusList2021Entry.fromJsonObject(credentialToValidate.vcDataModel.credentialStatus.jsonObject)
+
+    val statusLisCredStatusPurpose: String? =
+      statusListCredential.vcDataModel.credentialSubject.jsonObject[STATUS_PURPOSE] as? String?
+
+    requireNotNull(statusListEntryValue.statusPurpose)
```
bradleydwyer, line 118: `requireNotNull` here doesn't seem right. Typical use would be for checking function preconditions (i.e. this will throw an `IllegalArgumentException`). Same for below.

bradleydwyer: Is there anything we'd expect a caller to be able to do with the information from the checks here? Would it make more sense to just return false rather than make callers handle true/false & exception paths?

nitro-neal: It would alert the caller that they credentials they are passing in are not valid.

it will be true if the cred vs the credstatuslist is revoked
it will be false if the cred vs the credstatuslist is not revoked

it will throw an error if these are malformed creds.

I will fix up the exceptions to be more clear on that

credentials/src/main/kotlin/web5/sdk/credentials/StatusListCredential.kt
```diff
@@ -0,0 +1,301 @@
...
+   * The function then checks if the given `credentialToValidate`'s status list index is present in the expanded status list derived from the fetched status list credential.
+   *
+   * Example:
+   * ```
+   * val isRevoked = validateCredentialInStatusList(credentialToCheck)
+   * ```
+   */
+  public suspend fun validateCredentialInStatusList(
+    credentialToValidate: VerifiableCredential,
+    httpClient: HttpClient? = null // default HTTP client but can be overridden
```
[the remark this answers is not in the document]
nitro-neal: I was doing something similar that @andresuribe87 has in another PR for did-ion. It makes it easy for mocking and allows more flexibility to the user so they can use any client they want.

I could go either way though, it is a bit strange..

To be clear: Are you suggesting that I remove the   httpClient: HttpClient? = null parameter and just create our default one?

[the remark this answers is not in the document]
bradleydwyer: `HttpClient` is a ktor class, so we're making ktor part of the public API for the library, which seems odd to me. It restricts our capacity to choose a different http client in the future (which should be an implementation detail for the library).

I understand the mocking/configuration angle though. Maybe there is a different way (e.g. expose specific things we expect should be configurable via a type we own - and for mocking to have specific visibility limiting public access).

I thought I had searched for other instances in other PRs (and I did) but just doubled checked and see similar in the `DidIonConfiguration` class.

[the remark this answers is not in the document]
bradleydwyer: Given ktor supports different engines, it's not as bad I guess. We could just leave it for now and revisit.

bradleydwyer (approved):

credentials/src/main/kotlin/web5/sdk/credentials/StatusListCredential.kt
```diff
@@ -0,0 +1,317 @@
...
+   * @return A [VerifiableCredential] instance.
+   *
+   * Example:
+   * ```
+   * val statusListCredential = StatusListCredential.create("http://example.com/statuslistcred/id123", "http://example.com/issuers/1", StatusPurpose.REVOCATION, listOf(vc1,vc2))
+   * ```
+   */
+  @Throws(RuntimeException::class)
+  public fun create(
+    statusListCredentialId: String,
```
andresuribe87, line 68: Would it make sense to validate that this is a `URI`? We do something similar in https://github.com/TBD54566975/web5-kt/blob/6bc48b81e15a9e7be6316eda0bccf518c14e335b/dids/src/main/kotlin/web5/sdk/dids/DidIonManager.kt#L284

credentials/src/main/kotlin/web5/sdk/credentials/StatusListCredential.kt
```diff
@@ -0,0 +1,317 @@
...
+    issuer: String,
```
andresuribe87, line 69: Similar to the above comment, should this be validated to be a resolvable DID (or similar)?

credentials/src/main/kotlin/web5/sdk/credentials/StatusListCredential.kt
```diff
@@ -0,0 +1,317 @@
...
+    statusPurpose: StatusPurpose,
+    issuedCredentials: List<VerifiableCredential>
```
andresuribe87, line 71:
```suggestion
    issuedCredentials: Iterable<VerifiableCredential>
```

credentials/src/main/kotlin/web5/sdk/credentials/StatusListCredential.kt
```diff
@@ -0,0 +1,317 @@
...
+ *
+ */
+public object StatusListCredential {
+  /**
+   * Create a [StatusListCredential] with a specific purpose, e.g., for revocation.
+   *
+   * @param statusListCredentialId The id used for the resolvable path to the status list credential [String].
+   * @param issuer The issuer URI of the credential, as a [String].
+   * @param statusPurpose The status purpose of the status list cred, eg: revocation, as a [StatusPurpose].
+   * @param issuedCredentials The credentials to be included in the status list credential, eg: revoked credentials, list of type [VerifiableCredential].
```
andresuribe87, line 58: Seems like the passed in credentials MUST have the `credentialStatus` property present, is that right? If so, would it make sense to add validation?

credentials/src/main/kotlin/web5/sdk/credentials/StatusListCredential.kt
```diff
@@ -0,0 +1,317 @@
...
+      .build()
+
+    return VerifiableCredential(vcDataModel)
+  }
+
+  /**
+   * Validates if a given credential is part of the status list represented by a [VerifiableCredential].
+   *
+   * @param credentialToValidate The [VerifiableCredential] to be validated against the status list.
+   * @param statusListCredential The [VerifiableCredential] representing the status list.
```
andresuribe87, line 106: Curious why you decided to pass in both credentials? An alternative can be to pass in the `credentialToValidate` only, and have this method retrieve the status list. What was the rationale to pick one vs the other?

credentials/src/main/kotlin/web5/sdk/credentials/StatusListCredential.kt
```diff
@@ -0,0 +1,317 @@
...
+  ): Boolean {
+    val statusListEntryValue: StatusList2021Entry =
+      StatusList2021Entry.fromJsonObject(credentialToValidate.vcDataModel.credentialStatus.jsonObject)
+
+    val statusListCredStatusPurpose: String? =
+      statusListCredential.vcDataModel.credentialSubject.jsonObject[STATUS_PURPOSE] as? String?
+
+    if (statusListEntryValue.statusPurpose == null) {
+      throw IllegalArgumentException("Status purpose in the credential to validate is null")
+    }
```
andresuribe87, line 128: Here and elsewhere, you can simplify as shown below
```suggestion
    require (statusListEntryValue.statusPurpose != null) {
      "Status purpose in the credential to validate is null"
    }
```

credentials/src/main/kotlin/web5/sdk/credentials/StatusListCredential.kt
```diff
@@ -0,0 +1,317 @@
...
+   * This function fetches the status list credential from a URL present in the `credentialToValidate`.
+   * It supports using either a user-provided `httpClient` or a default client when no client is passed in.
+   * The function then checks if the given `credentialToValidate`'s status list index is present in the expanded status list derived from the fetched status list credential.
+   *
+   * Example:
+   * ```
+   * val isRevoked = validateCredentialInStatusList(credentialToCheck)
+   * ```
+   */
+  public suspend fun validateCredentialInStatusList(
```
andresuribe87, line 168: We ended up deciding to have all methods be synchronous in the public APIs (at least for now). Could you make is such that this function isn't suspendable?

credentials/src/main/kotlin/web5/sdk/credentials/StatusListCredential.kt
```diff
@@ -0,0 +1,317 @@
...
+      throw Exception("decoding compressed bitstring", e)
+    }
+
+    val bitstringInputStream = ByteArrayInputStream(decoded)
+    val byteArrayOutputStream = ByteArrayOutputStream()
+
+    try {
+      GZIPInputStream(bitstringInputStream).use { it.copyTo(byteArrayOutputStream) }
+    } catch (e: Exception) {
+      throw Exception("unzipping status list bitstring using GZIP", e)
```
andresuribe87, line 304:
```suggestion
      throw RuntimeException("unzipping status list bitstring using GZIP", e)
```

credentials/src/main/kotlin/web5/sdk/credentials/StatusListCredential.kt
```diff
@@ -0,0 +1,317 @@
...
+      return validateCredentialInStatusList(credentialToValidate, statusListCredential)
+    } finally {
+      if (isDefaultClient) {
+        clientToUse.close()
+      }
+    }
+  }
+
+  private fun defaultHttpClient(): HttpClient {
+    return HttpClient(CIO)
```
andresuribe87, line 191: I think you'd probably want to set content negotiation with jackson, similar to https://github.com/TBD54566975/web5-kt/blob/6bc48b81e15a9e7be6316eda0bccf518c14e335b/dids/src/main/kotlin/web5/sdk/dids/DidIonManager.kt#L112

credentials/src/main/kotlin/web5/sdk/credentials/StatusListCredential.kt
nitro-neal: yup that is a good idea

credentials/src/main/kotlin/web5/sdk/credentials/StatusListCredential.kt
nitro-neal: yup we have this check in 
prepareCredentialsForStatusList
      requireNotNull(vc.vcDataModel.credentialStatus) { "no credential status found in credential" }

credentials/src/main/kotlin/web5/sdk/credentials/StatusListCredential.kt
nitro-neal: yup sounds good, I think I had this but there was chat to make it more explicit, I'll put it back to require though as I think its better syntax

credentials/src/main/kotlin/web5/sdk/credentials/StatusListCredential.kt
nitro-neal: I did both because the issuer may have these objects not hosted yet.

```
  public fun validateCredentialInStatusList(
    credentialToValidate: VerifiableCredential,
    statusListCredential: VerifiableCredential
  ): Boolean {
  ```
  
  ```
    public suspend fun validateCredentialInStatusList(
    credentialToValidate: VerifiableCredential,
    httpClient: HttpClient? = null // default HTTP client but can be overridden
  ): Boolean {
  ```

File: credentials/src/main/kotlin/web5/sdk/credentials/StatusListCredential.kt (hunk 9 above)
Comment (nitro-neal) [reply credentials/src/main/kotlin/web5/sdk/credentials/StatusListCredential.kt:168]: ok updated with runblocking

File: credentials/src/main/kotlin/web5/sdk/credentials/StatusListCredential.kt (hunk 4 above)
Comment (nitro-neal) [reply credentials/src/main/kotlin/web5/sdk/credentials/StatusListCredential.kt:69]: Added uri check and resolvable check

File: credentials/src/main/kotlin/web5/sdk/credentials/StatusListCredential.kt [hunk 12]
```diff
@@ -0,0 +1,345 @@
...
+   * Example:
+   * ```
+   * val isRevoked = validateCredentialInStatusList(credentialToCheck)
+   * ```
+   */
+  public fun validateCredentialInStatusList(
+    credentialToValidate: VerifiableCredential,
+    httpClient: HttpClient? = null // default HTTP client but can be overridden
+  ): Boolean {
+    return runBlocking {
```
andresuribe87, line 195: nit: block only around the suspendable bits. This makes debugging much easier.

credentials/src/main/kotlin/web5/sdk/credentials/StatusListCredential.kt
```diff
@@ -0,0 +1,345 @@
...
+    return HttpClient(CIO) {
+      install(ContentNegotiation) {
+        jackson { jacksonObjectMapper() }
+      }
+    }
+  }
+
+  private suspend fun HttpClient.fetchStatusListCredential(url: String): VerifiableCredential {
+    try {
+      val response: io.ktor.client.statement.HttpResponse = this.get(url)
```
andresuribe87, line 224: Could this be imported? 
```suggestion
      val response: HttpResponse = this.get(url)
```

credentials/src/main/kotlin/web5/sdk/credentials/StatusListCredential.kt
```diff
@@ -0,0 +1,345 @@
...
+      if (response.status.isSuccess()) {
+        val body = response.bodyAsText()
+        return VerifiableCredential.parseJwt(body)
+      } else {
+        throw ClientRequestException(response, "Failed to retrieve VerifiableCredentialType from $url")
+      }
+    } catch (e: ClientRequestException) {
```
andresuribe87, line 231: This is throwing catching an exception that was thrown in line 229. Curious what the motivation is behind doing so?

credentials/src/main/kotlin/web5/sdk/credentials/StatusListCredential.kt
```diff
@@ -0,0 +1,345 @@
...
+      throw Exception("Failed to fetch the status list credential due to a request error: ${e.message}", e)
```
andresuribe87, line 232: In other places, we've went with the approach of setting explicit exception types, instead of using the base `Exception` class. See the following example https://github.com/TBD54566975/web5-kt/blob/6bc48b81e15a9e7be6316eda0bccf518c14e335b/dids/src/main/kotlin/web5/sdk/dids/DidIonManager.kt#L349

credentials/src/main/kotlin/web5/sdk/credentials/StatusListCredential.kt
```diff
@@ -0,0 +1,345 @@
...
+  }
+
+  private suspend fun HttpClient.fetchStatusListCredential(url: String): VerifiableCredential {
+    try {
+      val response: io.ktor.client.statement.HttpResponse = this.get(url)
+      if (response.status.isSuccess()) {
+        val body = response.bodyAsText()
+        return VerifiableCredential.parseJwt(body)
+      } else {
+        throw ClientRequestException(response, "Failed to retrieve VerifiableCredentialType from $url")
```
andresuribe87, line 229: Would it make sense to reuse the following?  https://github.com/TBD54566975/web5-kt/blob/6bc48b81e15a9e7be6316eda0bccf518c14e335b/dids/src/main/kotlin/web5/sdk/dids/DidIonManager.kt#L340

andresuribe87 (approved):
Thanks - some minor nits here and there.

credentials/src/main/kotlin/web5/sdk/credentials/StatusListCredential.kt
nitro-neal: good call

credentials/src/main/kotlin/web5/sdk/credentials/StatusListCredential.kt
nitro-neal: yikez yea this is messy, I refactored

PR #85, opened 2023-10-16
Add better support for Java clients

bradleydwyer:
Add various annotations and subtle changes to better support Java clients using this library.


bradleydwyer:
Would like to push this through. JUnit references in Gradle won't make sense because I haven't included the related tests in this PR. Will submit separately.

common/build.gradle.kts
```diff
@@ -11,6 +11,11 @@ repositories {
 
 dependencies {
   testImplementation(kotlin("test"))
+  testImplementation("org.junit.jupiter:junit-jupiter:5.10.0")
+}
+
+tasks.test {
+  useJUnitPlatform()
```
andresuribe87, line 18: This is already present in https://github.com/TBD54566975/web5-kt/blob/main/build.gradle.kts#L94

Whats the intention behind adding it here?

common/build.gradle.kts
```diff
@@ -11,6 +11,11 @@ repositories {
 
 dependencies {
   testImplementation(kotlin("test"))
+  testImplementation("org.junit.jupiter:junit-jupiter:5.10.0")
```
andresuribe87, line 14: This is already present in https://github.com/TBD54566975/web5-kt/blob/main/build.gradle.kts#L46

Curious why it's needed here?

credentials/build.gradle.kts
```diff
@@ -22,4 +22,5 @@ dependencies {
   implementation("decentralized-identity:did-common-java:1.9.0")
 
   testImplementation(kotlin("test"))
+  testImplementation("org.junit.jupiter:junit-jupiter:5.10.0")
```
andresuribe87, line 25: Given this is present in https://github.com/TBD54566975/web5-kt/blob/main/build.gradle.kts#L46

Curious why you're adding it here?

dids/build.gradle.kts
```diff
@@ -24,4 +24,6 @@ dependencies {
   implementation("com.google.crypto.tink:tink:1.10.0")
   implementation("decentralized-identity:did-common-java:1.9.0")
   testImplementation(kotlin("test"))
+  testImplementation("org.junit.jupiter:junit-jupiter:5.10.0")
```
andresuribe87, line 27: Similar to all the questions above.

andresuribe87 (commented):
If this is the approach we're settled on, I can see that it can be easily forgotten. Is it possible to enforce this with a lint tool with an existing (or perhaps custom) lint rule?

andresuribe87 (approved):

ALRubinger:
@jiyoontbd Raising visibility of this; Java callers would benefit from these types of enhancements - it opens up Web5 (and tbDEX) to a wider audience of PFIs who are likely running Java backends

jiyoontbd:
@ALRubinger thanks for raising this! 

i do think this is worth the effort to be more inclusive of java friends, however i would like to close this PR for two reasons

1. this PR has sat stale for a while now so there's some merge conflicts
2. i'd like to be able to have a long term fix for this, through the use of detekt rules (we already have written and applied the detekt rule in place for adding `@JvmOverloads` https://github.com/TBD54566975/tbd-detekt-rules/issues/3)

it looks like this PR is trying to also implement the use of `@JvmStatic` and `@JvmField` - i've written up these issues in detekt rules repo.

https://github.com/TBD54566975/tbd-detekt-rules/issues/14
https://github.com/TBD54566975/tbd-detekt-rules/issues/15

cc @bradleydwyer

PR #86, opened 2023-10-16
Example ION creation with our http client interface. 

andresuribe87:
This is an example of what we can do to decouple http client implementations from the code that's using it.


mistermoe (commented):
i thought the point of `ktor` was to act as an abstract http client? for example, you can use `okhttp` with `ktor` or you can use ktor's default engine

andresuribe87:
> i thought the point of `ktor` was to act as an abstract http client? for example, you can use `okhttp` with `ktor` or you can use ktor's default engine

Yes, you are 100% correct. I just found https://api.ktor.io/ktor-client/ktor-client-okhttp/io.ktor.client.engine.okhttp/-ok-http/index.html

Closing

PR #87, opened 2023-10-17
Secp256k1 low-s

mistermoe:
Ensures that `Secp256k1.sign` always returns the low-s signature

> [!NOTE]
> It's **highly unlikely** that AWS KMS or Cloud KMS natively supports guaranteed low-s signatures. Neither KMS' support deterministic k defined in [RFC6979](https://datatracker.ietf.org/doc/html/rfc6979) either. 
> 
> It's worth mentioning that lack of cloud KMS support isn't a deal breaker as the math required to compute low-s from high-s does not require any private key material. However, lack of deterministic-k support may be a deal breaker specifically w.r.t. managing ION update and recovery keys in a cloud KMS if ION requires deterministic signatures

## Background
An `ECDSA` signature for a given message and private key is not strictly unique. Specifically, if `(r,s)` is a valid signature, then `(r, mod(-s, n))` is also a valid signature. 

This means there are two valid signatures for every message/private key pair: one with a "low" s value and one with a "high" s value. standardizing acceptance of only 1 of the 2 prevents signature malleability issues. 

Signature malleability is a notable concern in Bitcoin which introduced the low-s requirement for all signatures in version 0.11.1.

---

So, why does this matter?  potentially transaction malleability? - 

Transaction malleability is an issue where the transaction ID (`txid`) can change after the transaction has been created but before it's been added to the block. If someone modifies the signature of a transaction (e.g. by using the high-s counterpart), it changes the `txid` without invalidating the transaction. 

---
     
## Rando helpful stuff
* `n` - a large prime number that defines the maximum number of points that can be created by adding the base point, G, to itself repeatedly. The base point
* `G` - AKA generator point. a predefined point on an elliptic curve.

### DIFF
--- a/crypto/src/main/kotlin/web5/sdk/crypto/Secp256k1.kt
+++ b/crypto/src/main/kotlin/web5/sdk/crypto/Secp256k1.kt
@@ -218,7 +218,21 @@ public object Secp256k1 : KeyGenerator, Signer {
     val sha256 = MessageDigest.getInstance("SHA-256")
     val payloadDigest = sha256.digest(payload)
 
-    val (rBigint, sBigint) = signer.generateSignature(payloadDigest)
+    val (rBigint, initialSBigint) = signer.generateSignature(payloadDigest)
+
+    // ensure s is always in the bottom half of n.
+    // why? - An ECDSA signature for a given message and private key is not strictly unique. Specifically, if
+    //       (r,s) is a valid signature, then (r, mod(-s, n)) is also a valid signature. This means there
+    //       are two valid signatures for every message/private key pair: one with a "low" s value and one
+    //       with a "high" s value. standardizing acceptance of only 1 of the 2 prevents signature malleability
+    //       issues. Signature malleability is a notable concern in Bitcoin which introduced the low-s
+    //       requirement for all signatures in version 0.11.1.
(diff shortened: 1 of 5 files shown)


crypto/src/main/kotlin/web5/sdk/crypto/Secp256k1.kt
andresuribe87, line 1: Soem tests would be nice

andresuribe87 (approved):
Awesome stuff, thanks!

crypto/src/main/kotlin/web5/sdk/crypto/Secp256k1.kt
mistermoe: hoping to use [these vectors](https://github.com/paulmillr/noble-secp256k1/tree/main/test/vectors) for tests when i have the bandwidth to write them. will create an issue to increase secp256k1 test coverage

mistermoe:
@andresuribe87, i'll hold off on merging until you let me know whether we need a `public fun getLowS(signature: ByteArray): ByteArray` that takes the signature provided and returns its low-s counterpart. we should know after trying ION DID creation with `AwsKeyManager` a bunch of times

phoebe-lew (approved):

PR #91, opened 2023-10-20
Presentation Definition Test refactors

phoebe-lew:


tomdaffurn (approved):

credentials/src/test/kotlin/web5/sdk/credentials/TestData.kt
andresuribe87, line 1: Oh I meant actually having the JSON contents into a separate file with the .json extension, but this works!

andresuribe87 (approved):

PR #94, opened 2023-10-24
Fix null pointer in VC verify

tomdaffurn:
# Overview
Fix a null pointer that occurs when verifying a VC signed with an ION DID that has no assertionMethod

# How Has This Been Tested?
New unit test added

### DIFF
--- a/credentials/src/main/kotlin/web5/sdk/credentials/VerifiableCredential.kt
+++ b/credentials/src/main/kotlin/web5/sdk/credentials/VerifiableCredential.kt
@@ -224,14 +224,9 @@ public class VerifiableCredential(public val vcDataModel: VcDataModel) {
       // using a set for fast string comparison. DIDs can be lonnng.
       val verificationMethodIds = setOf(parsedDidUrl.didUrlString, "#${parsedDidUrl.fragment}")
       val assertionMethods = didResolutionResult.didDocument.assertionMethodVerificationMethodsDereferenced
-      var assertionMethod: VerificationMethod? = null
-
-      for (method in assertionMethods) {
-        val id = method.id.toString()
-        if (verificationMethodIds.contains(id)) {
-          assertionMethod = method
-          break
-        }
+      val assertionMethod = assertionMethods?.firstOrNull {
+        val id = it.id.toString()
+        verificationMethodIds.contains(id)
       }
 
       if (assertionMethod == null) {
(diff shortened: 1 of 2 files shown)


credentials/src/main/kotlin/web5/sdk/credentials/VerifiableCredential.kt
```diff
@@ -224,14 +224,9 @@ public class VerifiableCredential(public val vcDataModel: VcDataModel) {
       // using a set for fast string comparison. DIDs can be lonnng.
       val verificationMethodIds = setOf(parsedDidUrl.didUrlString, "#${parsedDidUrl.fragment}")
       val assertionMethods = didResolutionResult.didDocument.assertionMethodVerificationMethodsDereferenced
```
mistermoe, line 226: i think we can leave 227-234 as is and just change the following yeah?

```suggestion
      val assertionMethods = didResolutionResult.didDocument.assertionMethodVerificationMethodsDereferenced ?: listOf()
```

tomdaffurn: That works too. I do prefer the functional style over a `for` loop and `break` though

mistermoe:
```kt
val assertionMethod = assertionMethods?.firstOrNull {
  val id = it.id.toString()
  verificationMethodIds.contains(id)
}
```

this only takes the first assertion method into consideration though yeah?

credentials/src/main/kotlin/web5/sdk/credentials/VerifiableCredential.kt
andresuribe87, line 1: MInd adding a test that exercises this logic?

credentials/src/main/kotlin/web5/sdk/credentials/VerifiableCredential.kt
tomdaffurn: `firstOrNull` iterates over the `Iterable` and returns the first element matching the given predicate, or null if element was not found.

mistermoe: oh nice! i'm just a scrub

credentials/src/main/kotlin/web5/sdk/credentials/VerifiableCredential.kt
tomdaffurn: Fair point. Trying to write one now but it's tricky because I need an ION DID without an assertionMethod, and it's private key so I can sign a VC. Will need to wait until #77 merges I think

mistermoe (approved):
🏅 🍖 🧇

phoebe-lew (approved):

PR #95, opened 2023-10-24
Cred status list impl 

nitro-neal:
Small updates to cred status list impl

### DIFF
--- a/credentials/src/main/kotlin/web5/sdk/credentials/StatusListCredential.kt
+++ b/credentials/src/main/kotlin/web5/sdk/credentials/StatusListCredential.kt
@@ -8,6 +8,7 @@ import io.ktor.client.plugins.ClientRequestException
 import io.ktor.client.plugins.ResponseException
 import io.ktor.client.plugins.contentnegotiation.ContentNegotiation
 import io.ktor.client.request.get
+import io.ktor.client.statement.HttpResponse
 import io.ktor.client.statement.bodyAsText
 import io.ktor.http.isSuccess
 import io.ktor.serialization.jackson.jackson
@@ -192,23 +193,25 @@ public object StatusListCredential {
     credentialToValidate: VerifiableCredential,
     httpClient: HttpClient? = null // default HTTP client but can be overridden
   ): Boolean {
-    return runBlocking {
       var isDefaultClient = false
       val clientToUse = httpClient ?: defaultHttpClient().also { isDefaultClient = true }
 
       try {
         val statusListEntryValue: StatusList2021Entry =
           StatusList2021Entry.fromJsonObject(credentialToValidate.vcDataModel.credentialStatus.jsonObject)
-        val statusListCredential =
-          clientToUse.fetchStatusListCredential(statusListEntryValue.statusListCredential.toString())
+        val statusListCredential: VerifiableCredential
 
-        return@runBlocking validateCredentialInStatusList(credentialToValidate, statusListCredential)
+        runBlocking {
+          statusListCredential =
+            clientToUse.fetchStatusListCredential(statusListEntryValue.statusListCredential.toString())
+        }
+
+        return validateCredentialInStatusList(credentialToValidate, statusListCredential)
       } finally {
         if (isDefaultClient) {
           clientToUse.close()
         }
       }
-    }
   }
 
   private fun defaultHttpClient(): HttpClient {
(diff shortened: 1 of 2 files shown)


credentials/src/main/kotlin/web5/sdk/credentials/StatusListCredential.kt
```diff
@@ -220,18 +223,26 @@ public object StatusListCredential {
...
+    }
+
     try {
-      val response: io.ktor.client.statement.HttpResponse = this.get(url)
       if (response.status.isSuccess()) {
         val body = response.bodyAsText()
         return VerifiableCredential.parseJwt(body)
       } else {
-        throw ClientRequestException(response, "Failed to retrieve VerifiableCredentialType from $url")
+        throw ClientRequestException(
```
andresuribe87, line 239: Throwing inside a `try` block seems like a code smell to me. Is that intentional?

credentials/src/test/kotlin/web5/sdk/credentials/StatusListCredentialTest.kt
```diff
@@ -361,7 +395,7 @@ class StatusListCredentialTest {
   }
 
   @Test
-  fun `should asynchronously validate if a credential is in the status list using a mock HTTP client`() = runBlocking {
+  fun `should asynchronously validate if a credential is in the status list using a mock HTTP client`() {
```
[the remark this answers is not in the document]
nitro-neal: has to make an http GET to download the status list credential (vs passing in the status list credential object)

credentials/src/main/kotlin/web5/sdk/credentials/StatusListCredential.kt
```diff
@@ -220,18 +223,23 @@ public object StatusListCredential {
...
-      throw Exception("Failed to fetch the status list credential due to a request error: ${e.message}", e)
-    } catch (e: ResponseException) {
-      throw Exception("Failed to fetch the status list credential due to a response error: ${e.message}", e)
+      response = this.get(url)
+    } catch (e: Exception) {
+      throw RuntimeException("Failed to retrieve VerifiableCredentialType from $url", e)
+    }
+
+    require(response.status.isSuccess()) {
+      "Failed to retrieve VerifiableCredentialType from $url with status ${response.status}"
```
andresuribe87, line 235: This throws an IllegalArgumentException. Is that what you intended?

credentials/src/main/kotlin/web5/sdk/credentials/StatusListCredential.kt
```diff
@@ -220,18 +223,23 @@ public object StatusListCredential {
...
+    }
+
+    try {
```
andresuribe87, line 238: I wonder if this try/catch block could be removed to simplify the code.

credentials/src/main/kotlin/web5/sdk/credentials/StatusListCredential.kt
```diff
@@ -249,15 +257,15 @@ public object StatusListCredential {
...
 
       val statusListEntry: StatusList2021Entry =
         StatusList2021Entry.fromJsonObject(vc.vcDataModel.credentialStatus.jsonObject)
 
-      require(statusListEntry.statusPurpose == statusPurpose.toString().lowercase()) { "status purpose mismatch" }
+      require(statusListEntry.statusPurpose == statusPurpose.toString().lowercase()) { "Status purpose mismatch" }
 
       if (!duplicateSet.add(statusListEntry.statusListIndex)) {
-        throw IllegalArgumentException("duplicate entry found with index: ${statusListEntry.statusListIndex}")
+        throw IllegalArgumentException("Duplicate entry found with index: ${statusListEntry.statusListIndex}")
```
andresuribe87, line 268:
```suggestion
       "Duplicate entry found with index: ${statusListEntry.statusListIndex}"
```

credentials/src/test/kotlin/web5/sdk/credentials/StatusListCredentialTest.kt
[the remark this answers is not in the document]
andresuribe87: I was also confused by this. Asynchronously, to me, means that you would need to explicitly wait until the http GET downloads the credential. That doesn't seem to be what the test is doing. Furthermore, that implementation has a `runBlocking` bit, which actually makes the async call be blocking. 

My suggestion is to rename to `should validate if a credential is in the status list using a mock HTTP client`

andresuribe87 (commented):
Looks good. There's a number of lines that aren't covered by tests. Would it make sense to expand the tests so that they're covered?

PR #96, opened 2023-10-25
Refactor and harmonizing DID method options

andresuribe87:
# Overview
Refactored according to offline conversation.

# Description
This PR
* Removes the option of passing `updateKeyJwk` and `recoveryKeyJwk`. 
* Allows for passing verification methods that can be generated by the keyManager inside ion.
* `CommonOptions` contains the common options across different operations.
* Renamed to use `Did<Method>Api` pattern.
* Renamed Did in favor of `StatefulDid`.
* Harmonized usage across different did method implementations.
* Updated documentation. 

# How Has This Been Tested?
Unit tests still pass. More tests were added.


dids/src/main/kotlin/web5/sdk/dids/DidKey.kt
```diff
@@ -63,10 +63,22 @@ public class DidKey(uri: String, keyManager: KeyManager) : Did(uri, keyManager)
...
+}
 
-  public companion object : DidMethod<DidKey, CreateDidKeyOptions> {
+/**
+ * API for interacting with "did:key" DIDS.
+ *
+ * ### Usage Example:
+ * ```kotlin
+ * val keyManager = InMemoryKeyManager()
+ * val did = DidKeyApi.create()
```
tomdaffurn, line 76:
```suggestion
 * val did = DidKeyApi.create(keyManager)
```

dids/src/main/kotlin/web5/sdk/dids/DidMethod.kt
```diff
@@ -24,7 +24,7 @@ import web5.sdk.crypto.KeyManager
  * Implementers should adhere to the respective DID method specifications ensuring both compliance
  * and interoperability across different DID networks.
  */
-public abstract class Did(public val uri: String, public val keyManager: KeyManager) {
+public abstract class StatefulDid(public val uri: String, public val keyManager: KeyManager) {
```
tomdaffurn, line 27: What's the thinking behind "stateful"? Is is describing that key managers have state?

tomdaffurn (dismissed):

dids/src/main/kotlin/web5/sdk/dids/DidMethod.kt
jiyoontbd: i had the same q!

jiyoontbd: also does this imply there's such a thing as a stateless did as well

dids/src/main/kotlin/web5/sdk/dids/DidIon.kt
```diff
@@ -446,37 +451,22 @@ public sealed class DidIonManager(
...
-      Pair(
-        PublicKey(
-          id = verificationMethodId,
-          type = "JsonWebKey2020",
-          publicKeyJwk = verificationJwk,
-          purposes = listOf(PublicKeyPurpose.AUTHENTICATION, PublicKeyPurpose.ASSERTION_METHOD),
-        ),
-        alias
-      )
+  private fun maybeGeneratePublicKeysToAdd(options: CommonOptions?, keyManager: KeyManager) =
```
phoebe-lew, line 460: nit: I know `maybe` in function names is totally a java thing buuut it's a pet peeve of mine 😢

dids/src/main/kotlin/web5/sdk/dids/DidIon.kt
```diff
@@ -724,30 +737,122 @@ public class ResolutionException(msg: String) : RuntimeException(msg)
  */
 public data class KeyAliases(
   public val updateKeyAlias: String?,
-  public val verificationKeyAlias: String?,
+  public val verificationKeyAlias: List<String>,
```
phoebe-lew, line 740:
```suggestion
  public val verificationKeyAliases: List<String>,
```

Can be multiple?

dids/src/main/kotlin/web5/sdk/dids/DidKey.kt
```diff
@@ -45,15 +45,15 @@ public class CreateDidKeyOptions(
...
+ * @constructor Initializes a new instance of [StatefulDidKey] with the provided [uri] and [keyManager].
  *
  * ### Usage Example:
  * ```kotlin
  * val keyManager = InMemoryKeyManager()
  * val did = DidKey("did:key:example", keyManager)
  * ```
  */
-public class DidKey(uri: String, keyManager: KeyManager) : Did(uri, keyManager) {
+public class StatefulDidKey(uri: String, keyManager: KeyManager) : StatefulDid(uri, keyManager) {
```
phoebe-lew, line 56: 🙏

dids/src/main/kotlin/web5/sdk/dids/DidMethod.kt
phoebe-lew: Yeah, basically based on the did-core spec (https://www.w3.org/TR/did-core/#terminology), `DID` itself just refers to the identifier, but our concept of `Did` contained the key manager + other things. And it's not immediately clear if we just say `Did` that it wraps the key manager (which you can also use independently of did proper operations)

dids/src/main/kotlin/web5/sdk/dids/DidIon.kt
andresuribe87: Done.

Yes, it can be multiple. You may add multiple verificationMethods to the DID Document. Each one can be used for different purposes (i.e. for different verification relationships).

dids/src/main/kotlin/web5/sdk/dids/DidIon.kt
andresuribe87: Renamed to `publicKeysWithAliasesToAdd`. Maybe that's better?

dids/src/main/kotlin/web5/sdk/dids/DidIon.kt
```diff
@@ -70,43 +73,41 @@ public class DidIonConfiguration internal constructor(
...
+public fun DidIonApi(builderAction: DidIonConfiguration.() -> Unit): DidIonApi {
   val conf = DidIonConfiguration().apply(builderAction)
-  return DidIonManagerImpl(conf)
+  return DidIonApiImpl(conf)
 }
 
-/** [DidIonManager] is sealed, so we provide an impl so the constructor can be called. */
-private class DidIonManagerImpl(configuration: DidIonConfiguration) : DidIonManager(configuration)
+/** [DidIonApi] is sealed, so we provide an impl so the constructor can be called. */
+private class DidIonApiImpl(configuration: DidIonConfiguration) : DidIonApi(configuration)
```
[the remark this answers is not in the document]
andresuribe87: Good question. First, let me get at the motivation. I wanted the users of the SDK to experience the following behaviors:

1. Able to call `create` like a static method

```kt
val did = DidIonApi.create()
```

2. Able to override some defaults with the builder pattern.

```kt
val api = DidIonApi {
  ionHost = "https://mycustomionhost.com"
}
val did = api.create()
```

3. Disable calling constructors for `DidIonApi`, so it can only be created with the builder pattern in 2. The motivation behind this is to remove the optionality on how to create `DidIonApi` objects. 


For 1, you need to have some sort of companion object.

For 2, you need something like 

```kt
public fun DidIonApi(configurationBlock: DidIonConfiguration.() -> Unit): DidIonApi
```

No problem so far. But for 3, that's where this kicks in. As you correctly pointed out, removing the `sealed` keyword would allow users of the SDK to call `DidIonApi(configuration)`.

dids/src/main/kotlin/web5/sdk/dids/DidMethod.kt
tomdaffurn: So it's a DID but also we have access to the key material. A DID that we "own".

Stateful is a bit confusing for me. Makes me think it's mutable, or not thread safe. Can't think of an obviously better name though. Best I've got is `DidResource`

andresuribe87: Depending on the DID method associated with a did, the DID Document associated with it may be mutable. DID ION, DID Web, and DID DHT are all example where that's the case.

Because of this, they aren't thread safe either.

nitro-neal: StatefulDid is a buzzkill on the naming to me..
could we keep the name Did but have like a property that gives the legit did without stateful stuff :hmm:

mistermoe: `StatefulDid` would continue to be confusing to me if we hadn't had a long discussion earlier this week. I still think `Did` is the right way to go. not sure that people will be thinking about the definition of `Did` strictly as per the spec. 

`StatefulDid` makes me think the concept of a `StatelessDid` exists as well which would effectively just be the DID as a string.

mistermoe: i think very few people are thinking about dids strictly as per spec and the likelihood of confusion caused by using the word `Stateful` is higher than the likelihood of confusion caused by not including it

mistermoe:
I think i walked away from our discussion earlier this week with something a bit different in mind. trying to figure out the best way to convey

mistermoe:
alright @andresuribe87 lmk if i'm off the mark here:

I think both of us have different use-cases in mind and i think both are valid! I'm confident we can find a way to support both

My desired Api surface. Use-case: services creating and primarily using a _single_ DID ->
```kt
// creation
val did = DidIon.create(km, opts)

// resolution - 2 paths
val resolutionResult = did.resolve()
DidIon.resolve("did:ion:whatever")

// updating - 2 paths
did.update(opts)
DidIon.update("did:ion:whatever", km, opts)
  
  
// recovering - single path. 
DidIon.recover("did:ion:whatever", km, opts)

maybe two paths are possible? though it seems weird to be able to call `did.recover()`
  

// deleting - 2 paths
did.delete(opts)
DidIon.delete("did:ion:whatever", km, opts)
  
// instantiating pre-existing did
val did = DidIon("did:ion:whatever", km, opts)
```

the instance methods simply use the static methods under the hood.

Now, I believe what this is missing that you would like to have is primarily two features:
* being able to set the host once for all subsequent _static_ `DidIon.*` fn calls
* being able to set a keymanager once for all subsequent _static_ `DidIon.*` fn calls
  

if i understand correctly, the motivation behind wanting the two missing features is that you're thinking about usage scenarios where _multiple_ dids have to be managed (e.g. building an SSI Service)


are my assumptions correct? and more importantly, if we figure out how to support the two missing features without changing the api surface i've described, can we satisfy both use-cases?

andresuribe87:
@mistermoe I think we're close to something that's going to be very robust! 🙌 

> alright @andresuribe87 lmk if i'm off the mark here:
> 
> I think both of us have different use-cases in mind and i think both are valid! I'm confident we can find a way to support both
> 
> My desired Api surface. Use-case: services creating and primarily using a _single_ DID ->
> 
> ```kotlin
> // creation
> val did = DidIon.create(km, opts)
> 
> // resolution - 2 paths
> val resolutionResult = did.resolve()
> DidIon.resolve("did:ion:whatever")
> 
> // updating - 2 paths
> did.update(opts)

Does `updateKeyAlias` live inside the `opts` object? I view this as an important detail because I don't know of a way of determining what the update key (even the public key) is by only looking at the DID Document (do you know if it's possible?). I have similar concerns with other operations below.

> ```kt
> DidIon.update("did:ion:whatever", km, opts)
>   
>   
> // recovering - single path. 
> DidIon.recover("did:ion:whatever", km, opts)
> 
> maybe two paths are possible? though it seems weird to be able to call `did.recover()`
> ```

Out of curiousity, where do you see the differences between `recover` and `update` (i.e. what exactly is weird about it)? 

> ```kt
>   
> 
> // deleting - 2 paths
> did.delete(opts)
> DidIon.delete("did:ion:whatever", km, opts)
>   
> // instantiating pre-existing did
> val did = DidIon("did:ion:whatever", km, opts)
> ```
> 
> the instance methods simply use the static methods under the hood.

Whether methods are static or not, I think that's an implementation detail. I would suggest focusing more on the syntax that we want to support.

> 
> Now, I believe what this is missing that you would like to have is primarily two features:
> 
> * being able to set the host _once_ for all subsequent _static_ `DidIon.*` fn calls

Yes! And also the `httpClient` engine (or the client itself). For a long lived service which is doing resolution, there is a benefit of not creating a ktor `HttpClient` per request (see [here](https://ktor.io/docs/create-client.html#close-client:~:text=Note%20that%20creating%20HttpClient%20is%20not%20a%20cheap%20operation%2C%20and%20it%27s%20better%20to%20reuse%20its%20instance%20in%20the%20case%20of%20multiple%20requests)). I've seen this pattern in many languages, so I assume the same is true if we were to use `OkHttp`. 

> * being able to set a keymanager _once_ for all subsequent _static_ `DidIon.*` fn calls

Almost! I agree with the spirit, but I don't think we need to impose whether some methods are static or not. As I mentioned above, I would want to support the specific syntax.

Note that I didn't address that in this PR as we decided to punt in our last sync.

> 
> if i understand correctly, the motivation behind wanting the two missing features is that you're thinking about usage scenarios where _multiple_ dids have to be managed (e.g. building an SSI Service)

Yes! I think a wallet app would also want to manage multiple dids?

> 
> are my assumptions correct? and more importantly, if we figure out how to support the two missing features without changing the api surface i've described (can certainly be additive), can we satisfy both use-cases?

Most of the assumptions are correct, thanks for putting together this nice summary!

I have an additional desire to use names that reduce future confusion for the abstractions we're creating.

Finally, a totally separate concern is that I think we should be fostering the explicit practice of using separate `KeyManager` instances to store the `recovery` vs other keys. I don't think that should be optional. I didn't actually address that in this PR.

Would love to hear any alternatives to handle the above!

mistermoe:
> Does updateKeyAlias live inside the opts object? I view this as an important detail because I don't know of a way of determining what the update key (even the public key) is by only looking at the DID Document (do you know if it's possible?). I have similar concerns with other operations below.

yeah ideally as optional. just spitballing here: it seems like our ion implementation can just default to a predetermined alias if one isn't provided e.g. `${shortFormDid}:update` given that there's only 1 update key at any given time. same pattern for recovery. 

sounds like a detail people shouldn't have to worry about unless they explicitly want to. do you have any reservations or hesitations with that approach?

mistermoe:
> Out of curiousity, where do you see the differences between recover and update (i.e. what exactly is weird about it)?

oh this was just out of ignorance actually. i hadn't looked into what `recover` actually meant in the context of ion. but it makes sense now. prior to reading the sidetree spec, i thought "hmmm how would i ever have an instance of the did if i am trying to recover it? doesn't it mean i lost it?." so we can ignore that comment. 2 paths there works just as well

mistermoe:
> Yes! And also the httpClient engine (or the client itself). For a long lived service which is doing resolution, there is a benefit of not creating a ktor HttpClient per request (see [here](https://ktor.io/docs/create-client.html#close-client:~:text=Note%20that%20creating%20HttpClient%20is%20not%20a%20cheap%20operation%2C%20and%20it%27s%20better%20to%20reuse%20its%20instance%20in%20the%20case%20of%20multiple%20requests)). I've seen this pattern in many languages, so I assume the same is true if we were to use OkHttp.

yeah totally. i'd hope we don't instantiate an http client per request in any scenario. small nuance here is: creating one once and using it for every request doesn't necessitate that we allow devs to provide their own http clients. if we see a good reason for allowing people to provide their own client then happy to support it.

andresuribe87:
> > Does updateKeyAlias live inside the opts object? I view this as an important detail because I don't know of a way of determining what the update key (even the public key) is by only looking at the DID Document (do you know if it's possible?). I have similar concerns with other operations below.
> 
> yeah ideally as optional. just spitballing here: it seems like our ion implementation can just default to a predetermined alias if one isn't provided e.g. `${shortFormDid}:update` given that there's only 1 update key at any given time. same pattern for recovery.
> 
> sounds like a detail people shouldn't have to worry about unless they explicitly want to. do you have any reservations or hesitations with that approach?

I agree that they shouldn't worry about that. I tried to design it in that way, but ended up having really strong coupling.

 I think that can work as long as we're ok with overwriting a key in the key manager, and as long as we have the ability to provide a custom key id when generating a new key within the manager. Do you think that the `AwsKeyManager` would be able to implement those two features?

mistermoe:
> I think that can work as long as we're ok with overwriting a key in the key manager, and as long as we have the ability to provide a custom key id when generating a new key within the manager. Do you think that the AwsKeyManager would be able to implement those two features?

yep for sure. something that Tom and I had previously discussed and punted until there was a need which it seems like we have now!

mistermoe:
> Finally, a totally separate concern is that I think we should be fostering the explicit practice of using separate KeyManager instances to store the recovery vs other keys. I don't think that should be optional. I didn't actually address that in this PR.

yep totally on board with encouraging this. only question for you here is: is providing a separate key manager a _MUST_ for you or a _CAN_/_SHOULD_? where _CAN_/_SHOULD_ implies that the key manager used for all other keys will also be used for recovery unless a recovery specific key manager is supplied

andresuribe87:
> > Finally, a totally separate concern is that I think we should be fostering the explicit practice of using separate KeyManager instances to store the recovery vs other keys. I don't think that should be optional. I didn't actually address that in this PR.
> 
> yep totally on board with encouraging this. only question for you here is: is providing a separate key manager a _MUST_ for you or a _CAN_/_SHOULD_? where _CAN_/_SHOULD_ implies that the key manager used for all other keys will also be used for recovery unless a recovery specific key manager is supplied

For me it's a MUST. And if people know what they're doing , they can override to use the same :)

andresuribe87:
> > I think that can work as long as we're ok with overwriting a key in the key manager, and as long as we have the ability to provide a custom key id when generating a new key within the manager. Do you think that the AwsKeyManager would be able to implement those two features?
> 
> yep for sure. something that Tom and I had previously discussed and punted until there was a need which it seems like we have now!

Is there a way we could support a `latest` tag for a versioned key? That would be my preference. That way there is a nice trail.

andresuribe87:
> > Yes! And also the httpClient engine (or the client itself). For a long lived service which is doing resolution, there is a benefit of not creating a ktor HttpClient per request (see [here](https://ktor.io/docs/create-client.html#close-client:~:text=Note%20that%20creating%20HttpClient%20is%20not%20a%20cheap%20operation%2C%20and%20it%27s%20better%20to%20reuse%20its%20instance%20in%20the%20case%20of%20multiple%20requests)). I've seen this pattern in many languages, so I assume the same is true if we were to use OkHttp.
> 
> yeah totally. i'd hope we don't instantiate an http client per request in any scenario. small nuance here is: creating one once and using it for every request doesn't necessitate that we allow devs to provide their own http clients. if we see a good reason for allowing people to provide their own client then happy to support it.

You're right! They're separate concerns. I would like to allow for that flexibility though, because I can see users of the SDK wanting to override that. An example would be if they want to do custom logging and/or tracing.

dids/src/main/kotlin/web5/sdk/dids/DidMethod.kt
KendallWeihe: In accordance with the [spec terminology](https://www.w3.org/TR/did-core/#terminology) "DID" is technically just the "_globally unique persistent identifier_", which basically means the string-value `did:key:sdlgjhsgolwnglwjnrg`, but... all of the marketing material (for lack of a better term) is centered around "DID" which would require it being an abstraction which encapsulates everything else (keys, docs, resolution, etc) for the purposes of aligning public awareness ("marketing") with implementation.

decentralgabe: no stateful pls 😄 

if anything you might want to tag some DIDs as "self resolving" - which would just apply to did:key and did:jwk - but I'm not certain the value of that

decentralgabe (approved):

mistermoe:
This is great @andresuribe87 . let's merge this right after hack week

mistermoe (approved):
awesome work @andresuribe87

PR #98, opened 2023-10-25
Example of injecting KeyManager to DidMethodApi objects

andresuribe87:


mistermoe:
should we close this PR @andresuribe87 ?

PR #99, opened 2023-10-26
fix/secp256k1

mistermoe:
* Account for signatures less than 64 bytes
* add test
* bump version

### DIFF
--- a/crypto/src/main/kotlin/web5/sdk/crypto/Secp256k1.kt
+++ b/crypto/src/main/kotlin/web5/sdk/crypto/Secp256k1.kt
@@ -102,7 +102,7 @@ public object Secp256k1 : KeyGenerator, Signer {
    */
   public const val COMPRESSED_KEY_SIZE: Int = 33
 
-  public const val SIG_LEN: Int = 64
+  public const val SIG_SIZE: Int = 64
 
   /**
    * Range that defines the position of the X coordinate in an uncompressed public key byte array.
@@ -244,7 +244,7 @@ public object Secp256k1 : KeyGenerator, Signer {
     val sBigint = if (initialSBigint >= halfN) curveParams.n.subtract(initialSBigint) else initialSBigint
 
     // Secp256k1 signatures are always 64 bytes.
-    return rBigint.toFixedByteArray(SIG_LEN / 2) + sBigint.toFixedByteArray(SIG_LEN / 2)
+    return rBigint.toFixedByteArray(SIG_SIZE / 2) + sBigint.toFixedByteArray(SIG_SIZE / 2)
   }
 
   override fun verify(publicKey: JWK, signedPayload: ByteArray, signature: ByteArray, options: VerifyOptions?) {
--- a/crypto/src/main/kotlin/web5/sdk/crypto/Secp256k1.kt
+++ b/crypto/src/main/kotlin/web5/sdk/crypto/Secp256k1.kt
@@ -102,6 +102,8 @@ public object Secp256k1 : KeyGenerator, Signer {
    */
   public const val COMPRESSED_KEY_SIZE: Int = 33
 
+  public const val SIG_LEN: Int = 64
+
   /**
    * Range that defines the position of the X coordinate in an uncompressed public key byte array.
    *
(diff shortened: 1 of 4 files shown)


crypto/src/main/kotlin/web5/sdk/crypto/Secp256k1.kt
```diff
@@ -349,4 +344,34 @@ public object Secp256k1 : KeyGenerator, Signer {
...
+ * If the original byte array is shorter than the expected length, the function will pad the array with leading 0 bytes
+ * to reach the required size.
+ *
+ * This adjustment is particularly useful when dealing with cryptographic operations where a fixed size byte array
+ * is expected, ensuring consistency and correctness of the data format.
+ *
+ * @param size The expected fixed length of the resulting byte array.
+ * @return A [ByteArray] of size [size], representing the [BigInteger] value.
+ */
+private fun BigInteger.toFixedByteArray(size: Int): ByteArray {
```
KendallWeihe, line 368: TIL you can `private` a function outside of a class in Kotlin

crypto/src/test/kotlin/web5/sdk/crypto/Secp256k1Test.kt
```diff
@@ -53,6 +55,34 @@ class Secp256k1Test {
     val base64UrlEncodedSig2 = Convert(sig2).toBase64Url(padding = false)
 
     assertEquals(base64UrlEncodedSig1, base64UrlEncodedSig2)
+  }
+
+  @Test
+  fun fuzz() {
```
KendallWeihe, line 61:
```suggestion
  fun `can verify 10k random payloads`() {
```

or whatevs, not sure what the intent here is

mistermoe: point here is to generate a sufficient number of signatures (aka 2 big ints) across a large enough surface area to suss out any edge cases there may be with signature verification

phoebe-lew (approved):

crypto/src/test/kotlin/web5/sdk/crypto/Secp256k1Test.kt
mistermoe: cool with "pressure test signature verification" as the test name?

KendallWeihe: yup 👍

mistermoe: done - https://github.com/TBD54566975/web5-kt/commit/8049d78d49280d274a377d215016c9cff21d56f8

crypto/src/test/kotlin/web5/sdk/crypto/Secp256k1Test.kt
```diff
@@ -53,6 +55,34 @@ class Secp256k1Test {
...
+    repeat(10_000) {
+      val payload = generateRandomString(Random().nextInt(100))
 
+      try {
+        val sig1 = Secp256k1.sign(privateKey, payload.toByteArray())
+        Secp256k1.verify(publicKey, payload.toByteArray(), sig1)
+      } catch (e: SignatureException) {
+        println("($it) $e. Payload: $payload")
+        throw e
+      }
```
andresuribe87, line 85: Perhaps this would be simpler? 
```suggestion
    repeat(10_000) {
      val payload = Random.nextBytes(Random.nextInt(100))

      try {
        val sig1 = Secp256k1.sign(privateKey, payload)
        Secp256k1.verify(publicKey, payload, sig1)
      } catch (e: SignatureException) {
        val payloadString = Convert(payload).toBase64Url(false)
        println("($it) $e. Payload (base64url encoded): $payloadString")
        throw e
      }
```

andresuribe87 (approved):
lgtm, smol suggestion

Would have been awesome to identify a test that fails before, but passes now.

mistermoe:
> lgtm, smol suggestion
> 
> Would have been awesome to identify a test that fails before, but passes now.

good point @andresuribe87 . will keep this in mind for future PRs! less valuable now but for context:

we were running into intermittent unit test failures on the tbdex side w.r.t. signature verification e.g. 

```
OfferingTest > can parse offering from a json string() FAILED
    java.security.SignatureException: Invalid Signature
        at web5.sdk.crypto.Secp256k1.verify(Secp256k1.kt:280)
        at web5.sdk.crypto.Signer$DefaultImpls.verify$default(Signer.kt:115)
        at web5.sdk.crypto.Crypto.verify(Crypto.kt:144)
```

@tomdaffurn and i narrowed it down to secp256k1 and wrote the test that we added to this PR to consistently reproduce the issue. we used the failing test to track down and fix the bug. 

is this the type of context you were suggesting might be helpful?

crypto/src/test/kotlin/web5/sdk/crypto/Secp256k1Test.kt
mistermoe: awesome suggestion.

andresuribe87:
@mistermoe I meant more to include a test in this PR that always failed before your fix, but passes after the fix. Sounds like you kinda had it? No big deal. 

Context was super useful, thx!

PR #102, opened 2023-10-27
Implement PresentationDefinition field filter schema

phoebe-lew:


credentials/src/main/kotlin/web5/sdk/credentials/PresentationExchange.kt
```diff
@@ -20,40 +28,77 @@ public object PresentationExchange {
...
-            throw UnsupportedOperationException("Field Filter is not implemented")
-          }
-
-          if (field.path.any { path -> credential.getFieldByJsonPath(path) == null }) {
-            satisfied = false
-            break
-          }
-        }
-        return satisfied
+      .forEach { inputDescriptorWithFields ->
```
nitro-neal, line 50: Does this cover the scenario:
if there are no fields and only a path?

credentials/src/test/kotlin/web5/sdk/credentials/TestData.kt
```diff
@@ -1,34 +1,51 @@
...
-                                  "${'$'}.type[*]"
-                              ],
-                              "filter": {
-                                  "type": "string",
-                                  "pattern": "^SanctionCredential${'$'}"
-                              }
-                          }
-                      ]
-                  }
+const val TBDEX_PD = """
```
nitro-neal, line 3: Should add another test where there is no filter, only a path

nitro-neal (approved):
Good job! I would say add a few more tests to make sure it works with the old test too basically ( a pd without a filter, just a path)

credentials/src/main/kotlin/web5/sdk/credentials/PresentationExchange.kt
```diff
@@ -20,40 +28,77 @@ public object PresentationExchange {
...
+      val vcPayloadJson = JsonPath.parse(vcPayload.toString())
+        ?: throw PresentationExchangeError("Failed to parse VC $vcPayload as JsonNode")
+
+      val matchedFields = field.path.mapNotNull { path -> vcPayloadJson.read<JsonNode>(path) }
+      if (matchedFields.isEmpty()) {
+        throw PresentationExchangeError("Could not find matching field for path: ${field.path.joinToString()}")
+      }
+
+      when {
+        field.filterSchema != null -> matchedFields.any {vcSatisfiesFieldFilterSchema(it, field.filterSchema!!)}
```
phoebe-lew, line 78: matched field could be either an array or a single value.
if array, filterSchema needs to be applied to each item in the array. If any items satisfy the schema, return true
if single value, apply schema

phoebe-lew: matched field could also be an object

credentials/src/main/kotlin/web5/sdk/credentials/PresentationExchange.kt
phoebe-lew: it should via ln79 but will add a test :)

credentials/src/main/kotlin/web5/sdk/credentials/PresentationExchange.kt
```diff
@@ -20,40 +28,77 @@ public object PresentationExchange {
...
+    }
+  }
 }
+
+/**
+ * Custom error class for exceptions related to the Presentation Exchange.
+ *
+ * @param message The error message.
+ */
+public class PresentationExchangeError(message: String) : Error(message)
```
andresuribe87, line 104: Couldn't find where error is defined, is it an exception?

credentials/src/test/kotlin/web5/sdk/credentials/PresentationExchangeTest.kt
```diff
@@ -0,0 +1,43 @@
...
+import web5.sdk.dids.DidKey
+import kotlin.test.Test
+
+class PresentationExchangeTest {
+  private val keyManager = InMemoryKeyManager()
+  private val issuerDid = DidKey.create(keyManager)
+  private val holderDid = DidKey.create(keyManager)
+  private val jsonMapper: ObjectMapper = ObjectMapper()
+    .registerKotlinModule()
+    .findAndRegisterModules()
```
andresuribe87, line 19: Curious what this does?

credentials/src/test/kotlin/web5/sdk/credentials/PresentationExchangeTest.kt
```diff
@@ -0,0 +1,43 @@
...
+import web5.sdk.crypto.InMemoryKeyManager
+import web5.sdk.dids.DidKey
+import kotlin.test.Test
+
+class PresentationExchangeTest {
+  private val keyManager = InMemoryKeyManager()
+  private val issuerDid = DidKey.create(keyManager)
+  private val holderDid = DidKey.create(keyManager)
+  private val jsonMapper: ObjectMapper = ObjectMapper()
+    .registerKotlinModule()
```
andresuribe87, line 18: Curious what this does?

credentials/src/test/kotlin/web5/sdk/credentials/TestData.kt
andresuribe87, line 1: Personally, I think it's a better practice to commit the contents of `TBDEX_PD` into a separate file called `tbdex_pd.json`. Then you can load from disk and verify. This is advantageous because then we can start reusing test cases in other implementations. 

A similar comment applies to `VC_JWT`. 

What do you think?

credentials/src/main/kotlin/web5/sdk/credentials/PresentationExchange.kt
```diff
@@ -20,40 +28,77 @@ public object PresentationExchange {
...
    *
-   * @param presentationDefinition The Presentation Definition to be evaluated.
-   * @return `true` if the Presentation Definition is satisfied, `false` otherwise.
-   * @throws UnsupportedOperationException if certain features like Submission Requirements or Field Filters are not implemented.
+   * @param vcJwt The VC in JWT format.
+   * @param presentationDefinition The Presentation Definition to be satisfied.
    */
   public fun satisfiesPresentationDefinition(
-    credential: VerifiableCredential,
+    vcJwt: String,
```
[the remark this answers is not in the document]
nitro-neal: I its because ssome of the filter could be for the actual payload of the vc
$.iss is the correct way to filter by issuer not vc.issuer (this will be removed as per spec)

credentials/src/test/kotlin/web5/sdk/credentials/TestData.kt
phoebe-lew: sure, can do that

credentials/src/main/kotlin/web5/sdk/credentials/PresentationExchange.kt
phoebe-lew: `Error` is `java.lang.Error`, I'm actually not sure when to use error vs exception

credentials/src/main/kotlin/web5/sdk/credentials/PresentationExchange.kt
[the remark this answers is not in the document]
phoebe-lew: Context starting here: https://sq-tbd.slack.com/archives/C05SXUR39FE/p1698361441821219?thread_ts=1698337236.592429&cid=C05SXUR39FE

credentials/src/main/kotlin/web5/sdk/credentials/PresentationDefinition.kt
```diff
@@ -127,6 +127,7 @@ public class SubmissionRequirement(
 /**
  * Enumeration representing presentation rule options.
  */
+// TODO this does not serialize correctly but sub reqs not supported right now
```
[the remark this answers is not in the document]
phoebe-lew: Jackson does not know how to serialize/deserialize `enum class Rules` and `enum class Optionality` in a `SubmissionRequirement`, we probably need to write custom serializers/deserializers. However, the PEX impl throws an exception if `SubmissionRequirement` is present, so it doesn't seem worth writing the custom modules. 

Want to have a convo with @decentralgabe around whether we should drop `SubmissionRequirement` from our data model entirely. 

Can open an issue to track.

credentials/src/test/kotlin/web5/sdk/credentials/PresentationExchangeTest.kt
phoebe-lew: This one gets jackson's objectmapper to work nicely with kotlin

credentials/src/test/kotlin/web5/sdk/credentials/PresentationExchangeTest.kt
phoebe-lew: Actually don't need this anymore! If we define custom serialization/deserialization modules, this tells the object mapper to pick those up. But managed to just use annotations to get the PD to serialize.

credentials/src/main/kotlin/web5/sdk/credentials/PresentationDefinition.kt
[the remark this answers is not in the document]
decentralgabe: I'd just drop it for now - can add it back when/if needed

credentials/src/main/kotlin/web5/sdk/credentials/PresentationExchange.kt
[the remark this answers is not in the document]
decentralgabe: agree w/ the concern, but it's better to decouple the logic IMO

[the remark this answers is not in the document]
andresuribe87: Sounds like you're advocating for leaving as is. I remain cautious about the approach. Agree with the intent of decoupling verification logic from validation.

My proposal is to stick to the following convention:
* When a function is dealing with data that's assumed to be trusted, use the `VerifiableCredential` type.
* When a function is dealing with data that needs to be verified, use the secured representation (i.e. `vcJwt` or DI).

For this function, the data is assumed to be trusted already. Given that, it would fall to the first bucket. 
On the other hand, the `VerifiableCredential.verify` function would follow the second pattern.


Alternatively, we could define types to represent:
* A secured verifiable credential: `JwVerifiableCredential` and `DataIntegrityVerifiableCredential`.
* An unsecured verifiable credential `UnsecuredVerifiableCredential`.

This would make it very difficult for users to screw up. 

Either of these two would be more beneficial IMHO. Those are my suggestions, I'll leave it up to @phoebe-lew to decide whether they're useful or not.

credentials/src/main/kotlin/web5/sdk/credentials/PresentationExchange.kt
andresuribe87: From https://docs.oracle.com/javase%2F7%2Fdocs%2Fapi%2F%2F/java/lang/Error.html

> An Error is a subclass of Throwable that indicates serious problems that a reasonable application should not try to catch.

That sounds really scary 😱 

In other PRs, we've tried to stick subclassing `RuntimeException`, like what's below https://github.com/TBD54566975/web5-kt/blob/02ea4b31a28cf68a4a3f9a9abc745e3273ceee79/dids/src/main/kotlin/web5/sdk/dids/DidIonManager.kt#L720

credentials/src/test/kotlin/web5/sdk/credentials/PresentationExchangeTest.kt
andresuribe87: Interesting! First time i've seen this. Thanks!

credentials/src/test/kotlin/web5/sdk/credentials/PresentationExchangeTest.kt
```diff
@@ -0,0 +1,113 @@
...
+    .setSerializationInclusion(JsonInclude.Include.NON_NULL)
+  @Suppress("MaximumLineLength")
+  val sanctionsVcJwt =
+    "eyJhbGciOiJFZERTQSJ9.eyJpc3MiOiJkaWQ6a2V5Ono2TWtrdU5tSmF0ZUNUZXI1V0JycUhCVUM0YUM3TjlOV1NyTURKNmVkQXY1V0NmMiIsInN1YiI6ImRpZDprZXk6ejZNa2t1Tm1KYXRlQ1RlcjVXQnJxSEJVQzRhQzdOOU5XU3JNREo2ZWRBdjVXQ2YyIiwidmMiOnsiQGNvbnRleHQiOlsiaHR0cHM6Ly93d3cudzMub3JnLzIwMTgvY3JlZGVudGlhbHMvdjEiXSwiaWQiOiIxNjk4NDIyNDAxMzUyIiwidHlwZSI6WyJWZXJpZmlhYmxlQ3JlZGVudGlhbCIsIlNhbmN0aW9uc0NyZWRlbnRpYWwiXSwiaXNzdWVyIjoiZGlkOmtleTp6Nk1ra3VObUphdGVDVGVyNVdCcnFIQlVDNGFDN045TldTck1ESjZlZEF2NVdDZjIiLCJpc3N1YW5jZURhdGUiOiIyMDIzLTEwLTI3VDE2OjAwOjAxWiIsImNyZWRlbnRpYWxTdWJqZWN0Ijp7ImlkIjoiZGlkOmtleTp6Nk1ra3VObUphdGVDVGVyNVdCcnFIQlVDNGFDN045TldTck1ESjZlZEF2NVdDZjIiLCJiZWVwIjoiYm9vcCJ9fX0.Xhd9nDdkGarYFr6FP7wqsgj5CK3oGTfKU2LHNMvFIsvatgYlSucShDPI8uoeJ_G31uYPke-LJlRy-WVIhkudDg"
+
+
+  private fun readPd(path: String): String {
+    return File(path).readText().trimIndent()
```
andresuribe87, line 28:
```suggestion
    return File(path).readText()
```

credentials/src/test/kotlin/web5/sdk/credentials/PresentationExchangeTest.kt
```diff
@@ -0,0 +1,113 @@
...
+  private val holderDid = DidKey.create(keyManager)
+  private val jsonMapper: ObjectMapper = ObjectMapper()
+    .registerKotlinModule()
+    .setSerializationInclusion(JsonInclude.Include.NON_NULL)
+  @Suppress("MaximumLineLength")
+  val sanctionsVcJwt =
+    "eyJhbGciOiJFZERTQSJ9.eyJpc3MiOiJkaWQ6a2V5Ono2TWtrdU5tSmF0ZUNUZXI1V0JycUhCVUM0YUM3TjlOV1NyTURKNmVkQXY1V0NmMiIsInN1YiI6ImRpZDprZXk6ejZNa2t1Tm1KYXRlQ1RlcjVXQnJxSEJVQzRhQzdOOU5XU3JNREo2ZWRBdjVXQ2YyIiwidmMiOnsiQGNvbnRleHQiOlsiaHR0cHM6Ly93d3cudzMub3JnLzIwMTgvY3JlZGVudGlhbHMvdjEiXSwiaWQiOiIxNjk4NDIyNDAxMzUyIiwidHlwZSI6WyJWZXJpZmlhYmxlQ3JlZGVudGlhbCIsIlNhbmN0aW9uc0NyZWRlbnRpYWwiXSwiaXNzdWVyIjoiZGlkOmtleTp6Nk1ra3VObUphdGVDVGVyNVdCcnFIQlVDNGFDN045TldTck1ESjZlZEF2NVdDZjIiLCJpc3N1YW5jZURhdGUiOiIyMDIzLTEwLTI3VDE2OjAwOjAxWiIsImNyZWRlbnRpYWxTdWJqZWN0Ijp7ImlkIjoiZGlkOmtleTp6Nk1ra3VObUphdGVDVGVyNVdCcnFIQlVDNGFDN045TldTck1ESjZlZEF2NVdDZjIiLCJiZWVwIjoiYm9vcCJ9fX0.Xhd9nDdkGarYFr6FP7wqsgj5CK3oGTfKU2LHNMvFIsvatgYlSucShDPI8uoeJ_G31uYPke-LJlRy-WVIhkudDg"
+
+
+  private fun readPd(path: String): String {
```
andresuribe87, line 27:
```suggestion
  private fun readFile(path: String): String {
```

andresuribe87 (approved):
Awesome!

PR #103, opened 2023-10-27
Implements SD-JWT

andresuribe87:
# Overview
Implements a module that supports SD-JWT according to https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-05.htm

# Description
SD-JWT is becoming the de-factor standard. This PR implements the basic functionality in order to support SD-JWT from the perspective of the holder, the issuer, and the verifier. Simple documentation has been added in this iteration.  It's published as a separate module, and it's dependencies are kept lean on purpose, since this module is meant to be used independently from DIDs and VCs. 

Most of the design takes inspiration from the Nimbus library, where a JWT needs to be created first, and then signed.

Feedback would be much appreciated. 

# How Has This Been Tested?
Unit tests everywhere.


sdjwt/src/main/kotlin/web5/security/SdJwt.kt
```diff
@@ -0,0 +1,364 @@
...
+    }
+    // Validate the signature over the SD-JWT.
+    // Validate the Issuer of the SD-JWT and that the signing key belongs to this Issuer.
+    val verifier = issuerVerifier(verificationOptions)
+    if (!issuerJwt.verify(verifier)) {
+      throw SignatureException("Verifying the issuerJwt failed: ${issuerJwt.serialize()}")
+    }
+
+    // Check that the SD-JWT is valid using nbf, iat, and exp claims, if provided in the SD-JWT, and not selectively disclosed.
+    val claimsVerifier = DefaultJWTClaimsVerifier<SecurityContext>(null, null)
```
nitro-neal, line 123: Are these nulls in the constructor expected?

sdjwt/src/main/kotlin/web5/security/SdJwt.kt
```diff
@@ -0,0 +1,364 @@
...
+
+    if (verificationOptions.holderBindingOption == HolderBindingOption.VerifyHolderBinding) {
+      // If Holder Binding JWT is not provided, the Verifier MUST reject the Presentation.
+      require(keyBindingJwt != null) {
+        "Holder binding required, but holder binding JWT not found"
+      }
+
+      // Ensure that a signing algorithm was used that was deemed secure for the application. Refer to [RFC8725], Sections 3.1
+      // and 3.2 for details.
+      require(verificationOptions.supportedAlgorithms.contains(keyBindingJwt.header.algorithm)) {
```
nitro-neal, line 134: did similar supportedAlgorithms above, maybe pull out into a fuction

sdjwt/src/main/kotlin/web5/security/SdJwt.kt
```diff
@@ -0,0 +1,364 @@
...
+        "the algorithm of keyBindingJwt (${keyBindingJwt.header.algorithm}) is not part of the declared list of " +
+          "supported algorithms (${verificationOptions.supportedAlgorithms})"
+      }
+      // The none algorithm MUST NOT be accepted.
+      require(keyBindingJwt.header.algorithm != JWSAlgorithm.NONE) {
```
nitro-neal, line 139: Same here

sdjwt/src/main/kotlin/web5/security/SdJwt.kt
```diff
@@ -0,0 +1,364 @@
...
+    public fun parse(encodedDisclosure: String): Disclosure {
+      // Decode the base64-encoded disclosure
+      val disclosureJson = Base64URL(encodedDisclosure).decodeToString()
+
+      // Parse the disclosure JSON into a list of elements
+      val disclosureElems = defaultMapper.readValue(disclosureJson, List::class.java)
+
+      // Ensure that the disclosure is object or array disclosure
+      when (disclosureElems.size) {
+        3 -> {
```
nitro-neal, line 317: fancy fancy

sdjwt/src/main/kotlin/web5/security/SdJwt.kt
```diff
@@ -0,0 +1,364 @@
...
+        2 -> {
+          // Create a Disclosure instance
+          return ArrayDisclosure(
+            salt = disclosureElems[0] as String,
+            claimValue = disclosureElems[1] as Any,
+            raw = encodedDisclosure
+          )
+        }
+
+        else -> throw IllegalArgumentException("Disclosure \"$encodedDisclosure\" must have exactly 3 elements")
```
nitro-neal, line 340: Is this correct? Seems like there is a case to handle 2 elements

sdjwt/src/main/kotlin/web5/security/SdJwtBlinder.kt
```diff
@@ -0,0 +1,306 @@
...
+              allDisclosures.addAll(disclosures)
+              val arrayDisclosures = disclosures.map {
+                mapOf(
+                  blindedArrayKey to it.digest(sdAlg)
+                )
+              }.toMutableList()
+
+              repeat(totalDigests(arrayDisclosures.size) - arrayDisclosures.size) {
+                val randBytes = ByteArray(32)
+                SecureRandom().nextBytes(randBytes)
```
nitro-neal, line 167: We so need external audits on this stuff :hahaha:

sdjwt/src/main/kotlin/web5/security/SdJwtBlinder.kt
```diff
@@ -0,0 +1,306 @@
...
+
+/** RecursiveBlindOption implements https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-04.html#name-option-3-sd-jwt-with-recurs */
+public object RecursiveBlindOption : BlindOption()
+
+/** Implements https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-05.html#name-array-elements */
+public object ArrayBlindOption : BlindOption()
+
+
+/** Interface for generating salt values as described in https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-05.html#disclosures_for_object_properties */
+public interface ISaltGenerator {
```
nitro-neal, line 52: Is there not a library for this? Is this a special case sd-jwt salt?

sdjwt/src/main/kotlin/web5/security/SdJwtBlinder.kt
```diff
@@ -0,0 +1,306 @@
...
+    blindedClaimsMap[sdAlgClaimName] = hash.ianaName
+
+    val payload = JWTClaimsSet.parse(blindedClaimsMap)
+
+    return SdJwt.Builder(jwtClaimsSet = payload, disclosures = disclosures)
+  }
+
+}
+
+private fun getNextPowerOfTwo(n: Int): Int {
```
nitro-neal, line 286: wow some leetcode stuff

sdjwt/src/test/kotlin/web5/security/SdJwtTest.kt
```diff
@@ -0,0 +1,196 @@
...
+      "given_name": "Mister",
+      "family_name": "Tee",
+      "birthdate": "1940-10-31",
+      "nationalities": ["US", "DE", "PO"],
+      "address": {
+        "street": "happy street 123",
+        "zip_code": "12345"
+      }
+    }""".trimIndent()
+    val claimsToBlind = mapOf(
```
nitro-neal, line 65: Why can't this be dynamic? Is it to give the option to the user? 
If they provide null as claimsToBlind can it dynamically do default blinding to all?

nitro-neal (commented):
Wow mind blown. Great stuff! I need to reread the sd-jwt spec and look over one more time.

Will this only be used for credentials? Does it make sense to put it in a sub package of credentials or something?

There will be more changes in the credentials package to make it user friendly to get SD JWT creds along with the normal cred they made?

VerifiableCredential.create(sdJwt=ture) // something like this?

sdjwt/src/test/kotlin/web5/security/SdJwtTest.kt
andresuribe87: The issuer of an SdJwt must, at some point, select which claims they'll allow the holder to selectively disclose. For those, they must also say how they'll be disclosable, since there are various mechanisms.

What do you think should be the way in which they are all blinded by default?

andresuribe87:
> Will this only be used for credentials?

No, it can be used for other types credentials, that aren't w3c vcs.

> There will be more changes in the credentials package to make it user friendly to get SD JWT creds along with the normal cred they made?

Yes, but that comes later. Once we have support for VC2.0, then we'll need to allow securing it using sd-jwt. That will likely be the piece that does belong inside the `credentials` module.

> VerifiableCredential.create(sdJwt=ture) // something like this?

Something like that sounds reasonable.

sdjwt/src/main/kotlin/web5/security/SdJwt.kt
```diff
@@ -0,0 +1,364 @@
...
+  @Synchronized
+  public fun signKeyBinding(signer: JWSSigner) {
+    keyBindingJwt?.sign(signer)
+  }
+
+  /**
+   * TODO: only accept a subset of algos for verification.
+   * Verifies this SD-JWT according to [verificationOptions].
+   */
+  @Throws(Exception::class)
```
nitro-neal, line 101: We should be doing Runetime exceptions like we do in the VC package

sdjwt/src/main/kotlin/web5/security/SdJwtBlinder.kt
```diff
@@ -0,0 +1,306 @@
...
+          allDisclosures.addAll(ds)
+        }
+
+        else -> blinded.add(elem)
+      }
+    }
+    return Pair(blinded, allDisclosures)
+  }
+
+  fun toBlindedClaimsAndDisclosures(
```
nitro-neal, line 107: Can we add documentation on these functions?

sdjwt/src/test/kotlin/web5/security/SdJwtBlinderTest.kt
```diff
@@ -0,0 +1,333 @@
...
+      "birthdate" to FlatBlindOption,
+      "updated_at" to FlatBlindOption,
+      "nationalities" to ArrayBlindOption,
+    )
+
+    val sdJwt = signer.blind(claims, claimsToBlind)
+
+    val expected = """{
+      "_sd": [
+        "CrQe7S5kqBAHt-nMYXgc6bdt2SH5aTY1sU_M-PgkjPI",
```
nitro-neal, line 99: have you tested its the same output as ssi-sdk?

nitro-neal (approved):
I looked through it one more time, but wow its a lot to understand.

Looks good to me, no red flags, but I'll pull it down and trace through unit tests locally

sdjwt/src/README.md
```diff
@@ -0,0 +1,39 @@
...
+See the test named `whole flow from issuer to holder to verifier` from [this file](./test/kotlin/web5/security/SdJwtTest.kt).
+You can run it by cloning this repo and running gradle as shown below:
+
+```shell
+go clone github.com/TBD54566975/web5-kt.git
+cd web5-kt/
+./gradlew sdjwt:test
+```
+
+## API Reference
```
nitro-neal, line 37: can add to this repo or the main VC repo a simple example on how to create an SD vc

sdjwt/src/main/kotlin/web5/security/SdJwt.kt
```diff
@@ -0,0 +1,366 @@
...
+  public fun verify(
```
nitro-neal, line 103: Maybe add some more detailed explanations on these comments, similar to our other packages, and add example code if it is short and sweet

sdjwt/src/main/kotlin/web5/security/SdJwtUnblinder.kt
```diff
@@ -0,0 +1,216 @@
...
+   * Unblinds [serializedSdJwt]. Follows the algorithm specified in https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-05.html#section-6.1
+   *
+   * An unblinded [JWTClaimsSet], where the hidden values in the `issuerJwt` are replaced with values from the
+   * `disclosures` presented.
+   */
+  @Throws(Exception::class)
+  public fun unblind(
+    serializedSdJwt: String
+  ): JWTClaimsSet {
+    // 2. Separate the Presentation into the SD-JWT, the Disclosures (if any), and the Holder Binding JWT (if provided).
```
nitro-neal, line 71:
```suggestion
    // Separate the Presentation into the SD-JWT, the Disclosures (if any), and the Holder Binding JWT (if provided).
```

sdjwt/src/main/kotlin/web5/security/SdJwtUnblinder.kt
```diff
@@ -0,0 +1,216 @@
...
+    val sdJwt = SdJwt.parse(serializedSdJwt)
+
+    // Check that the _sd_alg claim value is understood and the hash algorithm is deemed secure.
+    val hashAlg = sdJwt.issuerJwt.jwtClaimsSet.getHashAlg()
+
+    // For each Disclosure provided, calculate the digest over the base64url-encoded string as described in Section 5.1.1.2.
+    val disclosuresByDigest = sdJwt.disclosures.associateBy { it.digest(hashAlg) }
+
+    // Process the Disclosures and _sd keys in the SD-JWT as follows:
+    //
```
nitro-neal, line 81:
```suggestion

```

sdjwt/src/test/kotlin/web5/security/SdJwtBlinderTest.kt
```diff
@@ -0,0 +1,333 @@
...
+class SdJwtBlinderTest {
+
+  private val mapper = jacksonObjectMapper().apply {
+    enable(SerializationFeature.INDENT_OUTPUT)
+    setSerializationInclusion(JsonInclude.Include.NON_NULL)
+    setDefaultPrettyPrinter(CustomPrettyPrinter())
+  }
+
+  @Test
+  fun testExample1() {
```
nitro-neal, line 21: fix the test names in this file

sdjwt/src/test/kotlin/web5/security/SdJwtBlinderTest.kt
```diff
@@ -0,0 +1,333 @@
...
+      "birthdate": "1940-01-01",
+      "given_name": "John",
+      "nationalities": [
+        "US",
+        "DE"
+      ]
+    }""".trimIndent()
+
+    val keyManager = InMemoryKeyManager()
+    DidKey.create(keyManager)
```
nitro-neal, line 57: are these lines needed?

settings.gradle.kts
```diff
@@ -1,3 +1,4 @@
 rootProject.name = "web5"
 include("common", "crypto", "dids", "credentials")
```
nitro-neal, line 2:
```suggestion
include("common", "crypto", "dids", "credentials", "sdjwt")
```

nitro-neal:
Why is this another package? 

In the future should we abstract like a lot of this away and have something like:

```
data class StreetCredibility(val localRespect: String, val legit: Boolean)

val vc = VerifiableCredential.create(
  type = "StreetCred",
  issuer = "did:example:issuer",
  subject = "did:example:subject",
  data = StreetCredibility(localRespect = "high", legit = true)
)

val sdVcs = vc.toSdVcs()
```

Am I thinking about this the wrong way?

nitro-neal (approved):
Happy to get this in since it is isolated in another package, but I want to see how we fit this into the bigger picture of our VerifiableCredential packages.

Would love to the see the api interface easier to understand and/or integrated in our existing VC package if possible

sdjwt/src/main/kotlin/web5/security/SdJwt.kt
```diff
@@ -0,0 +1,366 @@
...
+    )
+    verifier.jcaContext.provider = BouncyCastleProviderSingleton.getInstance() as Provider
+    return verifier
+  }
+
+  private fun jwkToKey(jwk: JWK): Key {
+    return when (jwk) {
+      is ECKey -> jwk.toPublicKey()
+      is OctetKeyPair -> jwk.toPublicKey()
+      is OctetSequenceKey -> jwk.toSecretKey()
```
decentralgabe, line 186: why is this secret but the others are public?

sdjwt/src/main/kotlin/web5/security/SdJwt.kt
```diff
@@ -0,0 +1,366 @@
...
+          // Create a Disclosure instance
+          return ObjectDisclosure(
+            salt = disclosureElems[0] as String,
+            claimName = disclosureClaimName,
+            claimValue = disclosureElems[2] as Any,
+            raw = encodedDisclosure
+          )
+        }
+
+        2 -> {
```
decentralgabe, line 333: nit: put 2 before 3

sdjwt/src/main/kotlin/web5/security/SdJwt.kt
andresuribe87: Added more documentation, and an example.

sdjwt/src/main/kotlin/web5/security/SdJwt.kt
andresuribe87: I think we should migrate to a set of exceptions. Will update shortly.

sdjwt/src/main/kotlin/web5/security/SdJwt.kt
andresuribe87: Ooof, good catch. Removed this line. Should only be using asymmetric keys.

sdjwt/src/main/kotlin/web5/security/SdJwt.kt
andresuribe87: Done

sdjwt/src/main/kotlin/web5/security/SdJwtUnblinder.kt
andresuribe87: Done

sdjwt/src/main/kotlin/web5/security/SdJwtUnblinder.kt
andresuribe87: Done

sdjwt/src/test/kotlin/web5/security/SdJwtBlinderTest.kt
andresuribe87: They aren't! Removed.

sdjwt/src/test/kotlin/web5/security/SdJwtBlinderTest.kt
andresuribe87: Fixed, and added where they came from

sdjwt/src/test/kotlin/web5/security/SdJwtBlinderTest.kt
andresuribe87: The examples are coming from the spec itself, so we should be covered. In a later PR, I'll add loads of test vectors from the reference implementation.

settings.gradle.kts
andresuribe87: Actually, separated everything into single lines and sorted alphabetically.

sdjwt/src/main/kotlin/web5/security/SdJwtBlinder.kt
andresuribe87: Decided against it because they're all private funcs.

sdjwt/src/README.md
andresuribe87: Synced offline. The current docs seem sufficient.

decentralgabe (approved):
lgtm - would appreciate a review from @mistermoe or @phoebe-lew before merging

sdjwt/src/main/kotlin/web5/security/SdJwt.kt
```diff
@@ -0,0 +1,389 @@
...
+internal const val blindedArrayKey = "..."
+
+/**
+ * Represents a Selective Disclosure JWT as defined in https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-05.html#name-terms-and-definitions.
+ * A more detailed overview is available in https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-05.html#name-sd-jwt-structure.
+ *
+ * Note: While [issuerJwt] and [keyBindingJwt] are of type [SignedJWT], they may or may not be signed.
+ */
+public class SdJwt(
+  public val issuerJwt: SignedJWT,
```
phoebe-lew, line 35: So in order to use SD-JWT, if for example if I have VC in JWT format created by the `web5/credentials` package, do I first need to convert it into the nimbusds `SignedJWT` object?

sdjwt/src/main/kotlin/web5/security/SdJwt.kt
```diff
@@ -0,0 +1,389 @@
...
+ * The hash algorithm as described in https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-05.html#name-hash-function-claim
+ */
+public typealias HashFunc = (ByteArray) -> ByteArray
+
+internal val defaultMapper = jacksonObjectMapper()
+
+/**
+ * Represents a disclosure for an Object Property as defined in https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-05.html#name-disclosures-for-object-prop.
+ */
+public class ObjectDisclosure(
```
phoebe-lew, line 282: Maybe move out disclosure classes into separate file?

sdjwt/src/main/kotlin/web5/security/SdJwtBlinder.kt
phoebe-lew: @andresuribe87 can you add comments or doc for this function? For code readability rather than external documentation

sdjwt/src/main/kotlin/web5/security/SdJwtBlinder.kt
```diff
@@ -0,0 +1,310 @@
...
+  }
+
+}
+
+private fun getNextPowerOfTwo(n: Int): Int {
+  if (n <= 0) {
+    return 1
+  }
+
+  return 2.0.pow(ceil(log2((n + 1).toDouble())).toInt()).toInt()
```
phoebe-lew, line 295: comment would be helpful

sdjwt/src/test/kotlin/web5/security/SdJwtTest.kt
```diff
@@ -0,0 +1,196 @@
...
+      mapOf(
+        "street_address" to "123 Main St",
+        "locality" to "Anytown",
+      )
+    )
+    assertEquals("""{"street_address": "123 Main St", "locality": "Anytown"}""", result)
+  }
+
+  @Test
+  fun `whole flow from issuer to holder to verifier`() {
```
phoebe-lew, line 51: super helpful to understand how the whole thing works!

sdjwt/src/test/kotlin/web5/security/SdJwtTest.kt
```diff
@@ -0,0 +1,196 @@
...
+    val alias = keyManager.getDeterministicAlias(issuerPublicJwk)
+    val issuerSigner = KeyManagerSigner(keyManager, alias)
+
+    builder.issuerHeader = JWSHeader.Builder(jwsAlgorithm)
+      .type(JOSEObjectType.JWT)
+      .keyID(alias)
+      .build()
+    val sdJwt = builder.build()
+
+    sdJwt.signAsIssuer(issuerSigner)
```
phoebe-lew, line 95: Can I not sign this with my did?

sdjwt/src/main/kotlin/web5/security/SdJwt.kt
andresuribe87: Done.

PR #105, opened 2023-10-30
Add DID DHT Method support

decentralgabe:
Follows https://tbd54566975.github.io/did-dht-method

- Support for type indexing and two key types (ed25519, secp256k1)
- Support for packet encoding and decoding with tests
- Publishing to the DHT and resolution from the DHT to come in a follow-up


dids/src/main/kotlin/web5/sdk/dids/ion/model/Models.kt
```diff
@@ -18,6 +18,13 @@ import com.fasterxml.jackson.databind.ser.std.StdSerializer
 import com.nimbusds.jose.jwk.JWK
 import web5.sdk.common.Convert
 import web5.sdk.common.EncodingFormat
+import web5.sdk.dids.JacksonJwk
+import web5.sdk.dids.PublicKeyPurpose
+
+/**
+ * Represents an HTTP response where the status code is outside the range considered success.
+ */
+public class InvalidStatusException(public val statusCode: Int, msg: String) : RuntimeException(msg)
```
andresuribe87, line 27: This seems to belong in another package, as it's not ion specific.

andresuribe87 (commented):
Would suggest waiting until #96 is merged, as it changes the approach that's being taken across dids.

dids/build.gradle.kts
```diff
@@ -40,4 +40,7 @@ dependencies {
   testImplementation(kotlin("test"))
   testImplementation("io.ktor:ktor-client-mock:$ktor_version")
   testImplementation("org.mockito.kotlin:mockito-kotlin:5.1.0")
+
+  implementation("dnsjava:dnsjava:3.5.2")
+  implementation("io.github.m-m-m:mmm-binary:0.9.3")
```
decentralgabe, line 45: was the only open source z-base-32 lib I could find with an acceptable license that worked for the JVM. I tried chad but chad failed and failed again, this works.

dids/src/main/kotlin/web5/sdk/dids/Models.kt
```diff
@@ -0,0 +1,14 @@
+package web5.sdk.dids
+
+import com.fasterxml.jackson.annotation.JsonValue
+
+/**
+ * Enum representing the purpose of a public key.
+ */
+public enum class PublicKeyPurpose(@get:JsonValue public val code: String) {
```
decentralgabe, line 8: @andresuribe87 we should consider killing this since it's present in the DID lib we use

dids/build.gradle.kts
KendallWeihe: wow what a name `m-m-m.mmm-binary` 😆

dids/src/test/kotlin/web5/sdk/dids/DidDhtTest.kt
```diff
@@ -0,0 +1,165 @@
...
+      assertEquals(1, did.didDocument!!.capabilityInvocationVerificationMethods.size)
+      assertNull(did.didDocument!!.keyAgreementVerificationMethods)
+      assertNotNull(did.didDocument!!.services)
+      assertEquals(1, did.didDocument!!.services.size)
+      assertContains(did.didDocument!!.services[0].id.toString(), "test-service")
+    }
+  }
+
+  @Nested
+  inner class PacketTest {
```
KendallWeihe, line 92:
```suggestion
  inner class DnsPacketTest {
```

dids/src/test/kotlin/web5/sdk/dids/DidDhtTest.kt
```diff
@@ -0,0 +1,165 @@
...
+
+    @Test
+    fun `to and from DNS packet - simple DID`() {
```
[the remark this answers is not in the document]
decentralgabe: the DHT doesn't care about DNS at all. we are leveraging a project called [Pkarr](https://github.com/nuhvi/pkarr) which adds in the DNS semantics (worth reading!).

dids/src/main/kotlin/web5/sdk/dids/DidDht.kt
```diff
@@ -0,0 +1,420 @@
...
+ * Specifies options for creating a new "did:dht" Decentralized Identifier (DID).
+ * @property verificationMethodsToAdd A list of [JWK]s to add to the DID Document mapped to their purposes
+ * as verification methods.
+ * @property servicesToAdd A list of [Service]s to add to the DID Document.
+ * @property publish Whether to publish the DID Document to the DHT after creation.
+ */
+public class CreateDidDhtOptions(
+  public val verificationMethodsToAdd: Iterable<Pair<JWK, Array<PublicKeyPurpose>>>? = null,
+  public val servicesToAdd: Iterable<Service>? = null,
+  public val publish: Boolean? = false,
```
nitro-neal, line 46: wouldn't the default 80% case be to publish every time you make a new dht? Should we flip this default to true?

dids/src/main/kotlin/web5/sdk/dids/DidDht.kt
```diff
@@ -0,0 +1,420 @@
...
+          purposes.forEach { relationship ->
+            relationshipsMap.getOrPut(relationship) { mutableListOf() }.add(
+              VerificationMethod.builder().id(verificationMethod.id).build()
+            )
+          }
+        }
+      } ?: emptyList()) + identityVerificationMethod
+
+      // map to the DID object model's services
+      val services = opts.servicesToAdd?.map { service ->
```
nitro-neal, line 145: Should we have a require that makes sure all parts of the servicesToAdd are there?
(What if we have a null type or service endpoint?)

dids/src/main/kotlin/web5/sdk/dids/DidDht.kt
```diff
@@ -0,0 +1,420 @@
...
+          .publicKeyJwk(key.toPublicJWK().toJSONObject())
+          .build().also { verificationMethod ->
+          purposes.forEach { relationship ->
+            relationshipsMap.getOrPut(relationship) { mutableListOf() }.add(
+              VerificationMethod.builder().id(verificationMethod.id).build()
+            )
+          }
+        }
+      } ?: emptyList()) + identityVerificationMethod
+
```
nitro-neal, line 143:
```suggestion
      opts.servicesToAdd?.forEach { service ->
        requireNotNull(service.id) { "Service id cannot be null" }
        requireNotNull(service.type) { "Service type cannot be null" }
        requireNotNull(service.serviceEndpoint) { "Service serviceEndpoint cannot be null" }
      }
```

dids/src/main/kotlin/web5/sdk/dids/DidDht.kt
decentralgabe: yes, will add a note to do this when publishing is implemented

nitro-neal (approved):
LGTM after the build is fixed with the ZBase32 issue

dids/src/main/kotlin/web5/sdk/dids/DidDht.kt
```diff
@@ -0,0 +1,424 @@
...
+/**
+ * Specifies options for creating a new "did:dht" Decentralized Identifier (DID).
+ * @property verificationMethodsToAdd A list of [JWK]s to add to the DID Document mapped to their purposes
+ * as verification methods.
+ * @property servicesToAdd A list of [Service]s to add to the DID Document.
+ * @property publish Whether to publish the DID Document to the DHT after creation.
+ */
+public class CreateDidDhtOptions(
+  public val verificationMethodsToAdd: Iterable<Pair<JWK, Array<PublicKeyPurpose>>>? = null,
```
mistermoe, line 44: thoughts on `verificationMethods` instead of `verificationMethodsToAdd`? i feel like the `toAdd` is implied because it's `create`

dids/src/main/kotlin/web5/sdk/dids/DidDht.kt
```diff
@@ -0,0 +1,424 @@
...
+  public val servicesToAdd: Iterable<Service>? = null,
```
mistermoe, line 45: thoughts on `services` instead of `servicesToAdd`? i feel like the `toAdd` is implied because it's `create`

dids/src/main/kotlin/web5/sdk/dids/DidDht.kt
```diff
@@ -0,0 +1,424 @@
...
+      val keyAlias = keyManager.generatePrivateKey(JWSAlgorithm.EdDSA, Curve.Ed25519)
+      val publicKey = keyManager.getPublicKey(keyAlias)
+
+      // build DID Document
+      val id = getDidIdentifier(publicKey)
+
+      // add identity key to relationships map
+      val identityVerificationMethod =
+        VerificationMethod.builder()
+          .id(URI.create("$id#0"))
```
mistermoe, line 110: thoughts on using jwk thumbprint?

dids/src/main/kotlin/web5/sdk/dids/DidDht.kt
```diff
@@ -0,0 +1,424 @@
...
+    /**
+     * Converts a DNS packet to a [DIDDocument] according to the did:dht spec
+     * https://tbd54566975.github.io/did-dht-method/#dids-as-a-dns-packet
+     *
+     * @param did The DID that the packet is for.
+     * @param msg The [Message] instance containing the DNS packet.
+     * @return A [Pair] containing the [DIDDocument] and a list of types.
+     * @throws IllegalArgumentException if the provided DID does not conform to the "did:dht" method.
+     */
+    public fun fromDNSPacket(did: String, msg: Message): Pair<DIDDocument, List<Int>> {
```
mistermoe, line 320: 🙈 apologies in advance for saying: `fromDnsPacket` to remain consistent with convention in this repo

dids/src/main/kotlin/web5/sdk/dids/DidDht.kt
```diff
@@ -0,0 +1,424 @@
...
+    }
+
+    /**
+     * Checks whether a given DID identifier conforms to the "did:dht" method. This checks that the DID starts with
+     * "did:dht" and that the suffix is a valid z-base-32 encoded public key.
+     *
+     * @param did The DID to check.
+     * @return `true` if the DID conforms to the "did:dht" method, `false` otherwise.
+     */
+    public fun isValid(did: String): Boolean {
```
mistermoe, line 209: added alternative suggestion for this function [in this PR](https://github.com/TBD54566975/web5-kt/pull/115)

mistermoe (commented):
naisu @decentralgabe ! Added some comments / suggestions. Really excited about `did:dht`

dids/src/main/kotlin/web5/sdk/dids/DidDht.kt
decentralgabe: intentionally want to add a consistent first key to the DID document, thumbprint can be used for other keys in the doc

dids/src/main/kotlin/web5/sdk/dids/DidDht.kt
decentralgabe: I followed the naming convention in the rest of the project, you can see the ION create options are the same. I'm fine to rename here, but the others should be renamed as well.

dids/src/main/kotlin/web5/sdk/dids/DidDht.kt
decentralgabe: https://github.com/TBD54566975/web5-kt/pull/105#discussion_r1383759490

dids/src/main/kotlin/web5/sdk/dids/DidDht.kt
mistermoe: is the rationale behind why you want to do this documented somewhere? asking bc i know its going to come up over and over. would like to have a doc to point to

mistermoe (approved):

dids/src/main/kotlin/web5/sdk/dids/DidDht.kt
decentralgabe: no rationale (I can add) but it is specified here https://tbd54566975.github.io/did-dht-method/#create

the tl;dr on the rationale is:
- Have a guaranteed key to use (since it's the identity key, we can rely on it)
- Reduce size of the packet since 0 is shorter than a thumbprint

decentralgabe: https://github.com/TBD54566975/did-dht-method/issues/27#issuecomment-1795941823

mistermoe: awesome thanks!

PR #106, opened 2023-10-30
Support for DID web

andresuribe87:
# Overview
This PR adds support for resolution of DID who's method is web.

# Description
Note that creation wasn't added, as there are questions around how it should work. The main one,
* Given the SDK will not be able to modify the file in an external web domain, should the SDK facilitate creation and verification, similar to what https://github.com/TBD54566975/ssi-service/blob/main/doc/howto/didweb.md does?

# How Has This Been Tested?
Unit tests.


dids/src/main/kotlin/web5/sdk/dids/DidWeb.kt
```diff
@@ -0,0 +1,131 @@
...
+      for (i in 1 until subStrs.size) {
+        val str = URLDecoder.decode(subStrs[i], StandardCharsets.UTF_8)
+        urlBuilder.append("$str/")
+      }
+      urlBuilder.append(didDocFilename)
+      urlBuilder.toString()
+    }
+  }
+
+  public override fun create(keyManager: KeyManager, options: CreateDidOptions?): StatefulWebDid {
```
nitro-neal, line 125: We can create the json files needed for them to host  at their endpoints. (like this - https://identity.foundation/.well-known/did-configuration.json)

We should add this functionality eventually if not in this pr

dids/src/main/kotlin/web5/sdk/dids/DidWeb.kt
```diff
@@ -0,0 +1,131 @@
...
+ * @property uri The URI of the "did:web" which conforms to the DID standard.
+ * @property keyManager A [KeyManager] instance utilized to manage the cryptographic keys associated with the DID.
+ *
+ * ### Usage Example:
+ * ```kotlin
+ * val keyManager = InMemoryKeyManager()
+ * val did = StatefulWebDid("did:web:tbd.website", keyManager)
+ * ```
+ */
+public class StatefulWebDid(
```
nitro-neal, line 36: Can we just call it DidWeb? 
Even though it is a mutant object with a keyManager, I think its ok to call it didWeb and the didWeb.uri is the actual didWeb

dids/src/main/kotlin/web5/sdk/dids/DidWeb.kt
```diff
@@ -0,0 +1,131 @@
...
+    val parsedDid = DID.fromString(didWebStr)
+    require(parsedDid.methodName == methodName) {
+      "$didWebStr is missing prefix \"did:$methodName\""
+    }
+
+    val subStrs = parsedDid.methodSpecificId.split(":")
+
+    val decodedDomain = URLDecoder.decode(subStrs[0], StandardCharsets.UTF_8)
+
+    return if (subStrs.size == 1) {
```
nitro-neal, line 111: Can we get a unit test to cover this?

dids/src/main/kotlin/web5/sdk/dids/DidWeb.kt
```diff
@@ -0,0 +1,131 @@
...
+ */
+public fun DidWebApi(blockConfiguration: DidWebApiConfiguration.() -> Unit): DidWebApi {
+  val conf = DidWebApiConfiguration().apply(blockConfiguration)
+  return DidWebApiImpl(conf)
+}
+
+private class DidWebApiImpl(configuration: DidWebApiConfiguration) : DidWebApi(configuration)
+
+private const val wellKnownURLPath = ".well-known/"
+private const val didDocFilename = "did.json"
```
nitro-neal, line 61: where does did-configuration fit into this?

dids/src/main/kotlin/web5/sdk/dids/DidWeb.kt
andresuribe87: Done

dids/src/main/kotlin/web5/sdk/dids/DidWeb.kt
andresuribe87: Did you mean https://identity.foundation/.well-known/did.json ? If so, yes, I think that makes sense. That's similar to what we did in ssi-service

nitro-neal (approved):
awesome

dids/src/main/kotlin/web5/sdk/dids/DidWeb.kt
```diff
@@ -0,0 +1,127 @@
...
+
+    val url = URL(targetUrl.toString())
+    if (url.path.isEmpty()) {
+      targetUrl.append(wellKnownURLPath)
+    }
+    targetUrl.append(didDocFilename)
+    return targetUrl.toString()
+  }
+
+  public override fun create(keyManager: KeyManager, options: CreateDidOptions?): DidWeb {
```
nitro-neal, line 121: I think we should implement this so that users can create a did web doc similar to the way ssi service does it:

curl -X PUT https://ssi.tbddev.org/v1/dids/web \
-H "Content-Type: application/json" \
-d '{
  "keyType": "Ed25519",
  "options": {
    "didWebId": "did:web:example123.com"
  }
}'

nitro-neal:
Also, should you update the DidResolvers.kt file in this pr too?

dids/src/main/kotlin/web5/sdk/dids/DidWeb.kt
decentralgabe: I agree but think this sdk should be decoupled from a future http api

decentralgabe (approved):

PR #109, opened 2023-11-01
`dids` Package structure

mistermoe:
If you look at our autogenerate API reference documentation [here](https://tbd54566975.github.io/web5-kt/docs/htmlMultiModule/dids/web5.sdk.dids/index.html), it's crowded with `Ion` related stuff. 

Additionally, looking at the main page for the `dids` package [here](https://tbd54566975.github.io/web5-kt/docs/htmlMultiModule/dids/index.html), we have examples specific to `DidKey` immediately followed by ion stuff. 

This PR restructures the `dids` package a bit by introducing a `methods` directory (aka package). inside that package we can add 1 directory (aka package) per did method we support. This will make the api reference layout a bit cleaner and will also allow us to include examples in more relevant pages

@andresuribe87 i haven't moved the ion stuff. feel free to move that into `dids.methods.ion` if you're on board with this idea


andresuribe87:
@mistermoe makes sense to me! Moved the ion stuff into it's own module.

mistermoe:
@andresuribe87 nice! should we merge this branch into `did_alternatives`?

PR #115, opened 2023-11-04
`did:dht` validate

mistermoe:
@decentralgabe thoughts on refactoring `isValid` to `validate`? Rationale behind doing so would be:
* some of the function calls used in `isValid` could throw exceptions which means there are two possible outcomes that could occur if a did is invalid. Alternatively we could just try/catch return false in the function. mostly thinking that consistency might be nice
* thinking it may be nice to see why the did is invalid

Created this PR vs. using suggestion in your open PR because it would've ended up being like 3 disjointed suggestions.  Added tests for `validate` as well.


dids/src/main/kotlin/web5/sdk/dids/DidDht.kt
mistermoe, line 1: #96 is ✅  to merge, check out the updated directory structure there

decentralgabe (approved):

PR #136, opened 2023-11-14
Implement JWK with spec tests.

### LINKED ISSUE #129: Implement `did:jwk` creation & resolution
- [ ] Propose API surface design prior to implementation. This can be done by creating a draft PR and providing prototypal code or proposing the design in the comments of this issue

## Implementation Tasks:
- [ ] Produce/update API reference documentation for any/all methods added to the public API surface. documenting private API methods is optional
- [ ] Produce/update relevant example usage. add to `dids/module.md`
- [ ] Test coverage must be provided for all public API methods.
- [ ] Tests that consume shared test vectors must all be passing

andresuribe87:
# Overview
This PR implements `did:jwk` creation and resolution. It closes #129 

# Description
The API is the same as all other did methods. In particular, it's the same as `DidKey`. Documentation has also been updated. 

# How Has This Been Tested?
Implemented the examples from the spec. There are additional test vectors in a json implementation in https://github.com/OR13/did-jwk/blob/main/src/signatures-test-vectors.json, but they currently don't match the specification. https://github.com/OR13/did-jwk/issues/4 has more details. 

## References
* https://github.com/quartzjer/did-jwk/blob/main/spec.md

### DIFF
--- a/dids/module.md
+++ b/dids/module.md
@@ -6,6 +6,8 @@
 
 ## Creation
 
+### Creating a DID Key
+
 ```kt
 package example
 
@@ -14,6 +16,18 @@ import web5.sdk.dids.methods.key.DidKey
 
 val keyManager = InMemoryKeyManager()
 val did = DidKey.create(keyManager)
+```
+
+### Creating a DID Jwk
+
+```kt
+package example
+
+import web5.sdk.crypto.InMemoryKeyManager
+import web5.sdk.dids.methods.jwk.DidJwk
+
+val keyManager = InMemoryKeyManager()
+val did = DidJwk.create(keyManager)
 ```
 
 ## Export / Import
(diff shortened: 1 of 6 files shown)


mistermoe (approved):
awesome work @andresuribe87 !

mistermoe:
@andresuribe87 i feel like the test vectors you have here form a good foundation for what we can move into `sdk-development`.

andresuribe87:
> @andresuribe87 i feel like the test vectors you have here form a good foundation for what we can move into `sdk-development`.

Absolutely!

>  though maybe our vectors should be specific to the DSA we intend to support.

Good point. Yeah I think that makes. I'll move only the relevant ones to `sdk-development` once this is merged.

PR #137, opened 2023-11-14
Add Snyk for Vulns and License checks

### LINKED ISSUE #119: Setup License Checks
We need to set up license checks for this repository.

### LINKED ISSUE #120: Setup Security Scanning
We need to set up security scanning (most likely Snyk) for this repository.

leordev:
- add security scanning to detect vulns
- add license check
- upload report to Github Code Scanning

Fix #119
Fix #120

### DIFF
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -1,35 +1,31 @@
-name: Security Check
+name: Security Scanning
 
 on:
-  push:
+  pull_request_target:
     branches:
       - main
 
-  pull_request:
+  push:
     branches:
-      - "*"
+      - main
+      - leordev/* # todo: remove before merging
+
+  # Run every day at 5am UTC
+  schedule:
+    - cron: "0 5 * * *"
 
   # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:
 
 jobs:
-  security-audit:
+  snyk:
     runs-on: ubuntu-latest
+
     steps:
-      - name: Checkout source
+      - name: Checkout
         uses: actions/checkout@v4
-
-      - uses: gradle/wrapper-validation-action@v1
-
-      - uses: actions/setup-java@v3
         with:
-          distribution: "corretto"
-          java-version: 11
-
-      - uses: actions/cache@v3
-        with:
-          path: ~/.gradle/caches
-          key: gradle-${{ runner.os }}-${{ hashFiles('**/*.gradle.kts') }}-${{ hashFiles('**/gradle/wrapper/gradle-wrapper.properties') }}-${{ hashFiles('**/versions.properties') }}
+          ref: ${{ github.head_ref }}
 
       - name: Install Snyk
         uses: snyk/actions/setup@master
@@ -44,3 +40,23 @@ jobs:
         if: always()
         with:
           sarif_file: snyk.sarif
+
+  fossa:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.head_ref }}
+
+      - name: Run FOSSA Scan
+        uses: fossas/fossa-action@main
+        with:
+          api-key: ${{ secrets.FOSSA_API_KEY }}
+
+      - name: Run FOSSA Test
+        uses: fossas/fossa-action@main
+        with:
+          api-key: ${{ secrets.FOSSA_API_KEY }}
+          run-tests: true
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -17,7 +17,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout source
-        uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab # v3.5.2
+        uses: actions/checkout@v4
 
       - uses: gradle/wrapper-validation-action@v1
 
@@ -32,7 +32,7 @@ jobs:
           key: gradle-${{ runner.os }}-${{ hashFiles('**/*.gradle.kts') }}-${{ hashFiles('**/gradle/wrapper/gradle-wrapper.properties') }}-${{ hashFiles('**/versions.properties') }}
 
       - name: Install Snyk
-        uses: snyk/actions/setup@3e2680e8df93a24b52d119b1305fb7cedc60ceae # latest master (no released tag)
+        uses: snyk/actions/setup@master
 
       - name: Snyk VULN and License Check Test
         run: snyk test --all-projects --sarif-file-output=snyk.sarif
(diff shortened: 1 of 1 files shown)


.github/workflows/security.yml
```diff
@@ -0,0 +1,46 @@
...
+      - name: Snyk VULN and License Check Test
+        run: snyk test --all-projects --sarif-file-output=snyk.sarif
+        env:
+          SNYK_TOKEN: ${{ secrets.SNYK_TOKEN }}
+
+      - name: Upload SARIF result to GitHub Code Scanning
+        uses: github/codeql-action/upload-sarif@74483a38d39275f33fcff5f35b679b5ca4a26a99 # v2.22.5
+        if: always()
+        with:
+          sarif_file: snyk.sarif
```
andresuribe87, line 46: This seems quite a bit different from what's documented in https://github.com/snyk/actions/tree/master/gradle-jdk11

Mind elaborating on the differences?

leordev: By using the `setup` action we can expand later on to run SAST (code scanning using `snyk code test`) and more flexibility to play around the snyk-cli as if it is running in your local machine.

Source: https://github.com/snyk/actions/tree/master/setup

.github/workflows/security.yml
```diff
@@ -0,0 +1,46 @@
...
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+jobs:
+  security-audit:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout source
+        uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab # v3.5.2
```
andresuribe87, line 20:
```suggestion
        uses: actions/checkout@v4
```

.github/workflows/security.yml
```diff
@@ -0,0 +1,46 @@
...
+
+      - uses: actions/setup-java@v3
+        with:
+          distribution: "corretto"
+          java-version: 11
+
+      - uses: actions/cache@v3
+        with:
+          path: ~/.gradle/caches
+          key: gradle-${{ runner.os }}-${{ hashFiles('**/*.gradle.kts') }}-${{ hashFiles('**/gradle/wrapper/gradle-wrapper.properties') }}-${{ hashFiles('**/versions.properties') }}
```
andresuribe87, line 32: We do similar things in ci.yml. Would it make sense to use yaml anchors?

.github/workflows/security.yml
```diff
@@ -0,0 +1,46 @@
...
+
+      - name: Install Snyk
+        uses: snyk/actions/setup@3e2680e8df93a24b52d119b1305fb7cedc60ceae # latest master (no released tag)
```
andresuribe87, line 35: In case there are vulns in the setup action.
```suggestion
        uses: snyk/actions/setup@master
```

.github/workflows/security.yml
```diff
@@ -0,0 +1,46 @@
...
+
+      - name: Snyk VULN and License Check Test
+        run: snyk test --all-projects --sarif-file-output=snyk.sarif
+        env:
+          SNYK_TOKEN: ${{ secrets.SNYK_TOKEN }}
+
+      - name: Upload SARIF result to GitHub Code Scanning
+        uses: github/codeql-action/upload-sarif@74483a38d39275f33fcff5f35b679b5ca4a26a99 # v2.22.5
```
andresuribe87, line 43: Curious why pin this to a commit?

.github/workflows/security.yml
```diff
@@ -0,0 +1,46 @@
...
+        if: always()
```
andresuribe87, line 44: For my understanding, what does this `if` do? And is it possible to remove it?

andresuribe87 (approved):
Awesome, some more nits.

.github/workflows/security.yml
leordev: It will always upload the SARIF artifact, even if there are license checks failure (without it this step would never be reached in the run... in other words, it will always attempt to upload the SARIF artifact)

.github/workflows/security.yml
leordev: We pin commits purposefully because we try to get deterministic builds (for a given commit sha, that triggered this build branch, we want the same results, though not always perfect)... More details in the TBD Security Supply Chain doc [here](https://docs.google.com/document/d/1r22jO93WO44SNHawyrFhNhvT04G1ay9m3uf37wi15H8/edit?pli=1#heading=h.fypobv6byrbg).

.github/workflows/security.yml
leordev: Still not supported it seems... https://github.com/actions/runner/issues/1182

.github/workflows/security.yml
leordev: (see below)

.github/workflows/security.yml
andresuribe87: The build will still be reproducible if you user `@master` or a tagged commit. If you fix that commit, then any vulnerability found in the build tool (i.e. in the upload-sarif) will not be updated. How would you mitigate that?

.github/workflows/security.yml
andresuribe87: Ah, that sucks :(

.github/workflows/security.yml
leordev: This is a good call. @ALRubinger and I discussed about this point yesterday, and what we actually want is to make sure that for a given Git commit sha in our source code that our generated artifacts (binaries, jars etc) will have the same sha too. 

To achieve this I was restricting beyond what was needed by pinning the versions of everything in the pipeline, whereas in reality only the versions of dependencies used in our SDK should be pinned.

This is to avoid not only inconsistent artifacts sha but also introducing unexpected behavior in our sdks for our clients,  introducing new vulns etc.

But yes, you are right, we will probably be benefitted when the supporting tools in our pipeline are upgraded!

[TBD Supply Chain Security doc](https://docs.google.com/document/d/1r22jO93WO44SNHawyrFhNhvT04G1ay9m3uf37wi15H8/edit#heading=h.u50axc8wrgxq) (check the Reproducible Builds and Locked Versions sections)

leordev, line 43:
```suggestion
      - name: Upload SARIF result to GitHub Code Scanning
        uses: github/codeql-action/upload-sarif@v2
```

andresuribe87 (approved):
Awesome stuff @leordev !

mistermoe (approved):

.github/workflows/security.yml
ALRubinger: Yes this case is a choice of "choose your poison":

* Pin the dependency: always get reliable results from the build
* Let the dependency upgrade: `upload-sarif` may update over time or work differently and break builds

**In this case I think `@master` is the right call**. It doesn't impact the binary artifact generated, may receive important updates over time, and the scanning done here relies on an external service that doesn't have a pinned version anyway. Props to both @andresuribe87 for catching and @leordev in protecting against version ranges.

Issue #57, opened 2023-10-04
Integration tests setup


andresuribe87:

nitro-neal:
these should be created here - https://github.com/TBD54566975/web5-spec

mistermoe:
can you provide some context about this issue @andresuribe87 ?

PR #118, opened 2023-11-07
did:dht publishing and resolution

decentralgabe:
- publishing and resolution for did dht
- followed the pattern established by did:ion
- separated dht/pkarr logic into its own class which is fully tested (proper BEP44 impl)
- have network tests - can mock these out if wanted

PR #140, opened 2023-11-16
Use okhttp and add DOH for didweb

### LINKED ISSUE #111: Use `okhttp` instead of `ktor`

andresuribe87:
# Overview
Fixes #111 by using okhttp as the ktor engine. 

# Description
This also adds a client with DNS over HTTP for DIDWeb. 

# How Has This Been Tested?
Existing tests pass.


dids/src/main/kotlin/web5/sdk/dids/DidWeb.kt
```diff
@@ -82,7 +88,18 @@ public sealed class DidWebApi(
 
   private val mapper = jacksonObjectMapper()
 
-  private val engine: HttpClientEngine = configuration.engine ?: CIO.create {}
+  private val engine: HttpClientEngine = configuration.engine ?: OkHttp.create {
+    val appCache = Cache(File("cacheDir", "okhttpcache"), 10 * 1024 * 1024)
+    val bootstrapClient = OkHttpClient.Builder().cache(appCache).build()
+
+    val dns = DnsOverHttps.Builder().client(bootstrapClient)
```
decentralgabe, line 95: why is this preferable? why choose 8.8.8.8 over 9.9.9.9 (quad9) or 1.1.1.1 (cloudflare)?

phoebe-lew:
+1 to using okhttp as it's what we're using elsewhere, @andresuribe87 can you summarize why DNS over HTTP is better?

mistermoe:
@andresuribe87 what are the benefits of using `ktor`'s `okhttp` adapter over vanilla `okhttp`?

andresuribe87:
> +1 to using okhttp as it's what we're using elsewhere, @andresuribe87 can you summarize why DNS over HTTP is better?

Good question @phoebe-lew; First, a correction that I missed: it's DNS over HTTPS (not over HTTP). It makes it so that DNS resolution happens in a secure and privacy preserving manner. Thus, preventing man in the middle attacks. See https://w3c-ccg.github.io/did-method-web/#dns-security-considerations for details.

andresuribe87:
> @andresuribe87 what are the benefits of using `ktor`'s `okhttp` adapter over vanilla `okhttp`?

@mistermoe good question. By using `ktor`, we allow users of the SDK to use whatever httpclient engine they use within their stack. While many wont touch it, other companies may want to have a custom, internal, client. Additionally, it keeps the door open to doing [multiplatform development](https://kotlinlang.org/docs/multiplatform-get-started.html). For example, if we support [Kotlin native](https://kotlinlang.org/docs/native-overview.html#why-kotlin-native) in this SDK, that could be another option for having a single codebase for Android, JVM, Linux, and iOS.

dids/src/main/kotlin/web5/sdk/dids/DidWeb.kt
andresuribe87: This wasn't intentionally chosen. I copy pasted the setup as shown in https://github.com/square/okhttp/tree/master/okhttp-dnsoverhttps

Do you have any recommendations?

andresuribe87:
Since CI was failing because of code coverage, I took to opportunity to set up integration testing. The configuration I followed tracked https://docs.gradle.org/current/userguide/java_testing.html#sec:configuring_java_integration_tests

I added a couple of very basic integration tests as well. 

Thus, this PR will also fix #57

dids/src/intTest/kotlin/web5/sdk/dids/DidIntegrationTest.kt
mistermoe, line 1: can we have our integration test directory structure match the unit test directory structure?

mistermoe:
i'm guessing we'll keep #111  open until we replace the `CIO` engine with `okhttp` across the entire project?

dids/src/intTest/kotlin/web5/sdk/dids/DidIntegrationTest.kt
andresuribe87: I intentionally didn't do that because I expected integration tests to span across multiple packages and directories. For instance, I expect to add a test that reuses the same client across all different didmethods. I'm waiting until #118  is merged in order to do that. 

Given the above, what would you suggest?

dids/src/main/kotlin/web5/sdk/dids/DidWeb.kt
decentralgabe: I would recommend quad9 for its strong privacy characteristics

andresuribe87:
> i'm guessing we'll keep #111 open until we replace the `CIO` engine with `okhttp` across the entire project?

@mistermoe yes, that's the idea. I double checked that I had removed all instances that currently exist in the project, but probably missed some. Are there instances you found that I can remove?

dids/src/main/kotlin/web5/sdk/dids/DidWeb.kt
andresuribe87: Done in [9459174](https://github.com/TBD54566975/web5-kt/pull/140/commits/9459174f3c1b317af0210ef301429c45e5c2e998)

decentralgabe (approved):
lgtm

docs/examples.md
```diff
@@ -18,17 +18,18 @@ val did = DidIon.create(InMemoryKeyManager())
...
 ```kotlin
 val keyManager = InMemoryKeyManager()
 val ionApi = DidIonApi {
   ionHost = "my_custom_ion_host"
-  engine = CIO.create {
-    maxConnectionsCount = 10
-    requestTimeout = 5.toDuration(DurationUnit.SECONDS).inWholeMilliseconds
+  engine = OkHttp.create {
+    preconfigured = OkHttpClient.Builder()
+      .connectTimeout(Duration.ofSeconds(4))
```
decentralgabe, line 31: any reason to change from 5 to 4?

andresuribe87: Not really, just an example :)

mistermoe (approved):

PR #143, opened 2023-11-20
Implementing satisfiesPresentationDefinition

nitro-neal:
## Implementing satisfiesPresentationDefinition function for PEX

Validates a list of Verifiable Credentials (VCs) against a specified Presentation Definition.

This function ensures that the provided VCs meet the criteria defined in the Presentation Definition.

It first checks for the presence of Submission Requirements in the definition and throws an exception if they exist (not supported

It does this by mapping the input descriptors in the presentation definition to the corresponding VCs. If the number of mapped descriptors does not match the required count, an error is thrown.

More detailed error messages are planned for the future with a Results object that will give warnings, and better details for errors

### DIFF
--- a/credentials/src/main/kotlin/web5/sdk/credentials/PresentationExchange.kt
+++ b/credentials/src/main/kotlin/web5/sdk/credentials/PresentationExchange.kt
@@ -1,6 +1,8 @@
 package web5.sdk.credentials
 
 import com.fasterxml.jackson.databind.JsonNode
+import com.fasterxml.jackson.databind.node.ObjectNode
+import com.fasterxml.jackson.module.kotlin.jacksonObjectMapper
 import com.networknt.schema.JsonSchema
 import com.nfeld.jsonpathkt.JsonPath
 import com.nfeld.jsonpathkt.extension.read
@@ -37,13 +39,13 @@ public object PresentationExchange {
    * as this feature is not implemented. Then, it maps the input descriptors in the presentation definition to the
    * corresponding VCs. If the number of mapped descriptors does not match the required count, an error is thrown.
    *
-   * @param vcJwts List of VCs in JWT format to validate.
+   * @param vcJwts Iterable of VCs in JWT format to validate.
    * @param presentationDefinition The Presentation Definition V2 object against which VCs are validated.
    * @throws UnsupportedOperationException If Submission Requirements are present in the definition.
    * @throws PresentationExchangeError If the number of input descriptors matched is less than required.
    */
   public fun satisfiesPresentationDefinition(
-    vcJwts: List<String>,
+    vcJwts: Iterable<String>,
     presentationDefinition: PresentationDefinitionV2
   ) {
     if (!presentationDefinition.submissionRequirements.isNullOrEmpty()) {
@@ -64,26 +66,15 @@ public object PresentationExchange {
   }
 
   private fun mapInputDescriptorsToVCs(
-    vcJwtList: List<String>,
+    vcJwtList: Iterable<String>,
     presentationDefinition: PresentationDefinitionV2
   ): Map<InputDescriptorV2, List<String>> {
-    val map = mutableMapOf<InputDescriptorV2, MutableList<String>>()
-
-    presentationDefinition.inputDescriptors.forEach { inputDescriptor ->
-      val satisfyingVCs = mutableListOf<String>()
-
-      vcJwtList.forEach { vcJwt ->
-        if (vcSatisfiesInputDescriptor(vcJwt, inputDescriptor)) {
-          satisfyingVCs.add(vcJwt)
-        }
+    return presentationDefinition.inputDescriptors.associateWith { inputDescriptor ->
+      val satisfyingVCs = vcJwtList.filter { vcJwt ->
+        vcSatisfiesInputDescriptor(vcJwt, inputDescriptor)
       }
-
-      if (satisfyingVCs.size > 0) {
-        map[inputDescriptor] = satisfyingVCs
-      }
-    }
-
-    return map
+      satisfyingVCs
+    }.filterValues { it.isNotEmpty() }
   }
 
   /**
@@ -147,11 +138,27 @@ public object PresentationExchange {
     for (fieldValue in matchedFields) {
       if (fieldValue.isArray() && fieldValue.any { valueSatisfiesFieldFilterSchema(it, schema) }) {
         return true
-      } else if (!fieldValue.isArray() && valueSatisfiesFieldFilterSchema(fieldValue, schema)) {
+      }
+
+      if (fieldValue.isArray() && schema.isExpectingArray() && valueSatisfiesFieldFilterSchema(fieldValue, schema)) {
+        return true
+      }
+
+      if (!fieldValue.isArray() && valueSatisfiesFieldFilterSchema(fieldValue, schema)) {
         return true
       }
     }
     return false
+  }
+
+  private fun JsonSchema.isExpectingArray(): Boolean {
+    val schemaNode: JsonNode = this.schemaNode
+    return if (schemaNode is ObjectNode) {
+      val typeNode = schemaNode.get("type")
+      typeNode != null && typeNode.asText() == "array"
+    } else {
+      false
+    }
   }
 
   /**
--- a/credentials/src/main/kotlin/web5/sdk/credentials/PresentationExchange.kt
+++ b/credentials/src/main/kotlin/web5/sdk/credentials/PresentationExchange.kt
@@ -62,6 +62,7 @@ public object PresentationExchange {
       )
     }
   }
+
   private fun mapInputDescriptorsToVCs(
     vcJwtList: List<String>,
     presentationDefinition: PresentationDefinitionV2
--- a/credentials/src/main/kotlin/web5/sdk/credentials/PresentationExchange.kt
+++ b/credentials/src/main/kotlin/web5/sdk/credentials/PresentationExchange.kt
@@ -30,12 +30,17 @@ public object PresentationExchange {
   }
 
   /**
-   * Validates if a Verifiable Credential JWT satisfies a Presentation Definition.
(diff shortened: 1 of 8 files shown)


credentials/src/main/kotlin/web5/sdk/credentials/PresentationExchange.kt
```diff
@@ -31,75 +30,128 @@ public object PresentationExchange {
...
+        }
+      }
+
+      if (satisfyingVCs.size > 0) {
+        map[inputDescriptor] = satisfyingVCs
       }
+    }
+
+    return map
   }
```
andresuribe87, line 87: Perhaps this is simpler?
```suggestion
  private fun mapInputDescriptorsToVCs(
    vcJwtList: Iterable<String>,
    presentationDefinition: PresentationDefinitionV2
  ): Map<InputDescriptorV2, List<String>> {
    return presentationDefinition.inputDescriptors.associateWith { inputDescriptor ->
      val satisfyingVCs = vcJwtList.filter { vcJwt ->
        vcSatisfiesInputDescriptor(vcJwt, inputDescriptor)
      }
      satisfyingVCs
    }.filterValues { it.isNotEmpty() }
  }
```

credentials/src/main/kotlin/web5/sdk/credentials/PresentationExchange.kt
```diff
@@ -31,75 +30,128 @@ public object PresentationExchange {
...
+    val vcPayloadJson = JsonPath.parse(vc.payload.toString())
+      ?: throw PresentationExchangeError("Failed to parse VC payload as JSON.")
 
-      val matchedFields = field.path.mapNotNull { path -> vcPayloadJson.read<JsonNode>(path) }
-      if (matchedFields.isEmpty()) {
-        throw PresentationExchangeError("Could not find matching field for path: ${field.path.joinToString()}")
-      }
+    // If the Input Descriptor has constraints and fields defined, evaluate them.
+    inputDescriptor.constraints?.fields?.let { fields ->
+      val requiredFields = fields.filter { field -> field.optional != true }
```
andresuribe87, line 114:
```suggestion
      val requiredFields = fields.filter { field -> !field.optional }
```

credentials/src/main/kotlin/web5/sdk/credentials/PresentationExchange.kt
```diff
@@ -31,75 +30,128 @@ public object PresentationExchange {
...
-      .forEach { inputDescriptorWithFields ->
-        validateInputDescriptorsWithFields(inputDescriptorWithFields, vc.payload)
+    val inputDescriptorToVcMap = mapInputDescriptorsToVCs(vcJwts, presentationDefinition)
+
+    if (inputDescriptorToVcMap.size != presentationDefinition.inputDescriptors.size) {
+      throw PresentationExchangeError(
+        "Missing input descriptors: The presentation definition requires " +
+          "${presentationDefinition.inputDescriptors.size} descriptors, but only " +
+          "${inputDescriptorToVcMap.size} were found. Check and provide the missing descriptors."
+      )
```
andresuribe87, line 62: It would be incredibly useful if the library would tell the user what input descriptors were missing. What do you think?

andresuribe87 (approved):
This is awesome stuff @nitro-neal !

Some small comments

credentials/src/main/kotlin/web5/sdk/credentials/PresentationExchange.kt
andresuribe87: If you're addressing later, then disregard this comment!

credentials/src/test/resources/pd_filter_array.json
```diff
@@ -12,10 +12,7 @@
             ],
             "filter": {
               "type": "string",
-              "contains": {
-                "type": "string",
-                "const": "StreetCred"
-              }
+              "pattern": ".*StreetCred.*"
```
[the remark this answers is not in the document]
mistermoe: +1

credentials/src/main/kotlin/web5/sdk/credentials/PresentationExchange.kt
```diff
@@ -31,75 +30,128 @@ public object PresentationExchange {
...
+   * corresponding VCs. If the number of mapped descriptors does not match the required count, an error is thrown.
+   *
+   * @param vcJwts List of VCs in JWT format to validate.
+   * @param presentationDefinition The Presentation Definition V2 object against which VCs are validated.
+   * @throws UnsupportedOperationException If Submission Requirements are present in the definition.
+   * @throws PresentationExchangeError If the number of input descriptors matched is less than required.
    */
   public fun satisfiesPresentationDefinition(
-    vcJwt: String,
+    vcJwts: List<String>,
```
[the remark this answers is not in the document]
mistermoe: +1

mistermoe:
@nitro-neal are there any nuances specifically with respect what is/isn't supported for `filter` that we should document?

mistermoe:
one of my gripes with PEX v2 is the conflation of `path` and `filter`. e.g.

```
{
  "path": [
    "$.type[*]"
  ],
  "filter": {
    "type": "string",
    "pattern": "^StreetCredential$"
  }
}
```

vs.

```
{
  "path": [
    "$.type"
  ],
  "filter": {
    "type": "array",
    "contains": {
      "type": "string",
      "const": "StreetCredential"
    }
  }
}
```

the 1st field constraint makes it such that `filter` has to be applied to each element in the array whereas the 2nd field constraint leans on json schema entirely (preferred)

have we implemented support for both or just one of the two? if the latter, which one do we support?

credentials/src/test/resources/pd_filter_array_multiple_input_descriptors.json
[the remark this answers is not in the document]
nitro-neal: Well these are not really test vectors, just PDs for testing. I have a test vectors one coming soon (tm)

credentials/src/test/resources/pd_filter_array.json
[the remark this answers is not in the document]
nitro-neal: It was not valid, contains should only be used for array
for type string its pattern
              
https://json-schema.org/understanding-json-schema/reference/array

This will be caught and prevented in the next PR with PD validation + test vectors

credentials/src/main/kotlin/web5/sdk/credentials/PresentationExchange.kt
nitro-neal: Yup addressing in the next PR

credentials/src/main/kotlin/web5/sdk/credentials/PresentationExchange.kt
andresuribe87: Fair enough! If you keep the old syntax, can you make sure that you aren't returning a `MutableList` object in the map? You can dynamically create immutable collections using the `buildList` pattern as shown in https://kotlinlang.org/docs/constructing-collections.html#concrete-type-constructors

nitro-neal:
@mistermoe 

-with a small change- It now has support for both cases

I added a unit test to make sure its functioning as intended

credentials/src/test/kotlin/web5/sdk/credentials/PresentationExchangeTest.kt
```diff
@@ -89,11 +109,119 @@ class PresentationExchangeTest {
       )
       val vcJwt = vc.sign(issuerDid)
 
-      assertDoesNotThrow { PresentationExchange.satisfiesPresentationDefinition(vcJwt, pd) }
+      assertDoesNotThrow { PresentationExchange.satisfiesPresentationDefinition(listOf(vcJwt), pd) }
+    }
+
+    @Test
+    fun `does not throw when VC satisfies PD with no filter dob filed constraint`() {
```
phoebe-lew, line 116:
```suggestion
    fun `does not throw when VC satisfies PD with no filter dob field constraint`() {
```

credentials/src/test/kotlin/web5/sdk/credentials/PresentationExchangeTest.kt
```diff
@@ -89,11 +109,119 @@ class PresentationExchangeTest {
...
+        subject = holderDid.uri,
+        data = DateOfBirth(dateOfBirth = "1/1/1111")
+      )
+
+      val vcJwt = vc.sign(issuerDid)
+      assertDoesNotThrow { PresentationExchange.satisfiesPresentationDefinition(listOf(vcJwt), pd) }
+    }
+
+    @Test
+    fun `does not throw when VC satisfies PD with no filter dob filed constraint and extra VC`() {
```
phoebe-lew, line 134:
```suggestion
    fun `does not throw when VC satisfies PD with no filter dob field constraint and extra VC`() {
```

credentials/src/test/resources/pd_filter_array.json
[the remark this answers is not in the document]
phoebe-lew: Ah I think this was originally expecting the value at the path to be a string array, which is why `contains` + `type: string` was used. 

```
 "$.vc.type[*]",
 "$.type[*]"
```

If the value at these paths is a single string and not a string array, then yeah we need pattern.

credentials/src/main/kotlin/web5/sdk/credentials/PresentationExchange.kt
```diff
@@ -31,74 +32,132 @@ public object PresentationExchange {
...
+          "${inputDescriptorToVcMap.size} were found. Check and provide the missing descriptors."
+      )
+    }
+  }
+
+  private fun mapInputDescriptorsToVCs(
+    vcJwtList: Iterable<String>,
+    presentationDefinition: PresentationDefinitionV2
+  ): Map<InputDescriptorV2, List<String>> {
+    return presentationDefinition.inputDescriptors.associateWith { inputDescriptor ->
```
phoebe-lew, line 72: have not used `associateWith` before, cool!

phoebe-lew (approved):

PR #146, opened 2023-11-21
Replace `verificationMethods` type in `did:dht` to use `VerificationMethodSpec`

andresuribe87:
# Overview
This PR makes the `CreateDid{Dht,Ion}Options` consistent, and allows users to create `dht` dids without having to worry about the key creation for the verification methods.

# Description
See the updated `DidDhtTest.kt` file to understand how we allow consumers to create `VerificationMethods` with or without a `KeyManager`. 

# How Has This Been Tested?
Simple refactor. I just moved stuff around and _slightly_ updated tests.

## References
See the original `did:ion` PR that allowed this in #96

### DIFF
--- a/dids/src/main/kotlin/web5/sdk/dids/methods/dht/DidDht.kt
+++ b/dids/src/main/kotlin/web5/sdk/dids/methods/dht/DidDht.kt
@@ -30,9 +30,8 @@ import web5.sdk.dids.DidResolutionResult
 import web5.sdk.dids.PublicKeyPurpose
 import web5.sdk.dids.ResolveDidOptions
 import web5.sdk.dids.validateKeyMaterialInsideKeyManager
-import web5.sdk.dids.verificationmethods.VerificationMethodGenerator
 import web5.sdk.dids.verificationmethods.VerificationMethodSpec
-import web5.sdk.dids.verificationmethods.toGenerators
+import web5.sdk.dids.verificationmethods.toPublicKeys
 import java.net.URI
 
 /**
@@ -148,16 +147,26 @@ public sealed class DidDhtApi(configuration: DidDhtConfiguration) : DidMethod<Di
     }
 
     // map to the DID object model's verification methods
-    val verificationMethods = opts.verificationMethodsToAdd
-      .toGenerators(keyManager)
-      .toVerificationMethods(id).map { (purposes, verificationMethod) ->
-        purposes.forEach { relationship ->
-          relationshipsMap.getOrPut(relationship) { mutableListOf() }.add(
-            VerificationMethod.builder().id(verificationMethod.id).build()
-          )
-        }
-        verificationMethod
-      } + identityVerificationMethod
+    val verificationMethodToPublicKey = opts.verificationMethodsToAdd
+      .toPublicKeys(keyManager)
+      .associate { (_, publicKey) ->
+        val key = publicKey.publicKeyJwk
+        val verificationMethod = VerificationMethod.builder()
+          .id(URI.create("$id#${publicKey.id}"))
+          .type(publicKey.type)
+          .controller(URI.create(id))
+          .publicKeyJwk(key.toPublicJWK().toJSONObject())
+          .build()
+        verificationMethod to publicKey
+      }
+    verificationMethodToPublicKey.forEach { (verificationMethod, publicKey) ->
+      publicKey.purposes.forEach { relationship ->
+        relationshipsMap.getOrPut(relationship) { mutableListOf() }.add(
+          VerificationMethod.builder().id(verificationMethod.id).build()
+        )
+      }
+    }
+    val verificationMethods = verificationMethodToPublicKey.keys + identityVerificationMethod
     opts.servicesToAdd.forEach { service ->
       requireNotNull(service.id) { "Service id cannot be null" }
       requireNotNull(service.type) { "Service type cannot be null" }
@@ -176,7 +185,7 @@ public sealed class DidDhtApi(configuration: DidDhtConfiguration) : DidMethod<Di
     val didDocument =
       DIDDocument.builder()
         .id(URI(id))
-        .verificationMethods(verificationMethods)
+        .verificationMethods(verificationMethods.toList())
         .services(services)
         .assertionMethodVerificationMethods(relationshipsMap[PublicKeyPurpose.ASSERTION_METHOD])
         .authenticationVerificationMethods(relationshipsMap[PublicKeyPurpose.AUTHENTICATION])
@@ -541,24 +550,6 @@ public sealed class DidDhtApi(configuration: DidDhtConfiguration) : DidMethod<Di
       val (key, value) = it.split("=")
       key to value
     }
-  }
-}
-
-private fun Iterable<VerificationMethodGenerator>.toVerificationMethods(id: String)
-  : List<Pair<Iterable<PublicKeyPurpose>, VerificationMethod>> {
-  return this.map {
-    val (_, publicKey) = it.generate()
-    val key = publicKey.publicKeyJwk
-    val purposes = publicKey.purposes
-    Pair(
-      purposes,
-      VerificationMethod.builder()
-        .id(URI.create("$id#${publicKey.id}"))
-        .type(publicKey.type)
-        .controller(URI.create(id))
-        .publicKeyJwk(key.toPublicJWK().toJSONObject())
-        .build()
-    )
   }
 }
 
--- a/dids/src/main/kotlin/web5/sdk/dids/methods/dht/DidDht.kt
+++ b/dids/src/main/kotlin/web5/sdk/dids/methods/dht/DidDht.kt
@@ -30,6 +30,9 @@ import web5.sdk.dids.DidResolutionResult
 import web5.sdk.dids.PublicKeyPurpose
 import web5.sdk.dids.ResolveDidOptions
 import web5.sdk.dids.validateKeyMaterialInsideKeyManager
+import web5.sdk.dids.verificationmethods.VerificationMethodGenerator
+import web5.sdk.dids.verificationmethods.VerificationMethodSpec
+import web5.sdk.dids.verificationmethods.toGenerators
 import java.net.URI
 
 /**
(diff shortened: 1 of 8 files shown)


dids/src/main/kotlin/web5/sdk/dids/methods/dht/DidDht.kt
```diff
@@ -146,27 +147,33 @@ public sealed class DidDhtApi(configuration: DidDhtConfiguration) : DidMethod<Di
...
-          }
-        }
-    } ?: emptyList()) + identityVerificationMethod
-    opts.services?.forEach { service ->
+    val verificationMethodToPublicKey = opts.verificationMethodsToAdd
+      .toPublicKeys(keyManager)
+      .associate { (_, publicKey) ->
+        val key = publicKey.publicKeyJwk
+        val verificationMethod = VerificationMethod.builder()
+          .id(URI.create("$id#${publicKey.id}"))
```
decentralgabe, line 155: we need to make sure there are no ids of `0` which would conflict with the identity key

dids/src/main/kotlin/web5/sdk/dids/verificationmethods/VerificationMethod.kt
```diff
@@ -0,0 +1,125 @@
...
+    return Pair(null, PublicKey(id, JsonWebKey2020Type, controller, publicKeyJwk, relationships))
+  }
+}
+
+/**
+ * A [VerificationMethodSpec] according to https://w3c-ccg.github.io/lds-ecdsa-secp256k1-2019/.
+ *
+ * The [id] property cannot be over 50 chars and must only use characters from the Base64URL character set.
+ */
+public class EcdsaSecp256k1VerificationKey2019VerificationMethod(
```
decentralgabe, line 64: I think we should axe all usage of this in favor of JWK

mistermoe:
@andresuribe87 does this change the public api surface?

dids/src/main/kotlin/web5/sdk/dids/methods/dht/DidDht.kt
```diff
@@ -77,14 +79,13 @@ private class DidDhtApiImpl(configuration: DidDhtConfiguration) : DidDhtApi(conf
...
- * @property services A list of [Service]s to add to the DID Document.
+ * @property verificationMethodsToAdd List of specs that will be added to the DID ION document.
+ * @property servicesToAdd A list of [Service]s to add to the DID Document.
  * @property publish Whether to publish the DID Document to the DHT after creation.
  */
 public class CreateDidDhtOptions(
-  public val verificationMethods: Iterable<Pair<JWK, Array<PublicKeyPurpose>>>? = null,
-  public val services: Iterable<Service>? = null,
+  public val verificationMethodsToAdd: Iterable<VerificationMethodSpec> = emptyList(),
+  public val servicesToAdd: Iterable<Service> = emptyList(),
```
mistermoe, line 88: `toAdd` seems redundant here given that these are _create_ options.

dids/src/main/kotlin/web5/sdk/dids/verificationmethods/VerificationMethod.kt
```diff
@@ -0,0 +1,125 @@
...
+
+private const val JsonWebKey2020Type = "JsonWebKey2020"
+
+/**
+ * A [VerificationMethodSpec] according to https://w3c-ccg.github.io/lds-jws2020/.
+ *
+ * The [id] property cannot be over 50 chars and must only use characters from the Base64URL character set.
+ */
+public class JsonWebKey2020VerificationMethod(
+  public val id: String,
```
mistermoe, line 49: i think this should be optional not required. default to jwk thumbprint if not provided

dids/src/main/kotlin/web5/sdk/dids/verificationmethods/VerificationMethod.kt
andresuribe87: Can you elaborate on the motivation?

As a library, I think we should allow users to choose what is best fit for their purposes, as long as supporting it is not a big maintenance burden.

dids/src/main/kotlin/web5/sdk/dids/methods/dht/DidDht.kt
andresuribe87: Done

dids/src/main/kotlin/web5/sdk/dids/methods/dht/DidDht.kt
mistermoe: i can create a separate issue for changing `DidIon`. `DidDht` was already just `services` and `verificationMethods`. We should keep it that way. no point in changing code to match what's going to be immediately changed

dids/src/main/kotlin/web5/sdk/dids/verificationmethods/VerificationMethod.kt
decentralgabe: in what case is it a better fit for purpose to use this instead of a jwk?

andresuribe87: I didn't want to guess about how other consumers would intend to use it. Thinking more about it, I can see that it's reasonable to not support it at all.

andresuribe87: This would result in changes to the test vectors, as some (like https://github.com/TBD54566975/web5-kt/blob/d967931830113e21ad667207feeb742e123c4458/dids/src/test/resources/publicKeyModel1.json#L2) are using this method. I'll open a separate issue to address this.

dids/src/main/kotlin/web5/sdk/dids/methods/dht/DidDht.kt
andresuribe87: Ah, I see what you mean. Fixed in [c5b2eac](https://github.com/TBD54566975/web5-kt/pull/146/commits/c5b2eac1b5447a56f0600d340151f8a2790fdc38)

andresuribe87: And renamed to the existing didion to [b24b9fd](https://github.com/TBD54566975/web5-kt/pull/146/commits/b24b9fdcbdb63ef1a41465381165c1154291ad0f)

dids/src/main/kotlin/web5/sdk/dids/verificationmethods/VerificationMethod.kt
decentralgabe: thanks!

decentralgabe (approved):

dids/src/main/kotlin/web5/sdk/dids/verificationmethods/VerificationMethod.kt
andresuribe87: Elaborating further, I don't think we should be adding for a couple of reasons. The `id` here is the verification method ID. It's not the id for the key. The key id for JWK is contained inside the `publicKeyJwk` parameter.

dids/src/main/kotlin/web5/sdk/dids/methods/dht/DidDht.kt
```diff
@@ -77,14 +79,14 @@ private class DidDhtApiImpl(configuration: DidDhtConfiguration) : DidDhtApi(conf
 
 /**
  * Specifies options for creating a new "did:dht" Decentralized Identifier (DID).
- * @property verificationMethods A list of [JWK]s to add to the DID Document mapped to their purposes
- * as verification methods.
+ * @property verificationMethods List of specs that will be added to the DID ION document. It's important to note
```
decentralgabe, line 82:
```suggestion
 * @property verificationMethods List of verification methods that will be added to the DID DHT document. It's important to note
```

dids/src/test/kotlin/web5/sdk/dids/methods/dht/DidDhtTest.kt
```diff
@@ -252,8 +290,12 @@ class DidDhtTest {
 
       val otherKey = manager.generatePrivateKey(JWSAlgorithm.ES256K, Curve.SECP256K1)
       val publicKeyJwk = manager.getPublicKey(otherKey).toPublicJWK()
-      val verificationMethodsToAdd: Iterable<Pair<JWK, Array<PublicKeyPurpose>>> = listOf(
-        Pair(publicKeyJwk, arrayOf(PublicKeyPurpose.AUTHENTICATION, PublicKeyPurpose.ASSERTION_METHOD))
+      val verificationMethodsToAdd = listOf(
+        JsonWebKey2020VerificationMethod(
```
decentralgabe, line 294: noting that we should change this to just `JsonWebKey` https://github.com/TBD54566975/did-dht-method/issues/60 https://github.com/TBD54566975/sdk-development/issues/73

decentralgabe (approved):

dids/src/test/kotlin/web5/sdk/dids/methods/dht/DidDhtTest.kt
```diff
@@ -97,14 +98,51 @@ class DidDhtTest {
...
+                    JWSAlgorithm.EdDSA,
+                    Curve.Ed25519
+                  )
+                ).toPublicJWK(),
+                relationships = listOf(PublicKeyPurpose.AUTHENTICATION)
+              )
+            ),
+            publish = false
+          )
+        )
```
mistermoe, line 123: i'm just not sure anyone would be able to figure out how to do this

andresuribe87: @mistermoe is there a concrete suggestion you have in mind?

This was my best approach towards having options such that:
- The options between ion and Dht are consistent.
- there is a way for consumers of the library to provide the JWK keys they want to be used within the did document. 
- there is a way for consumers to have the key creation handled for them. 

Note that this had been discussed beforehand, when we added options to did ion. It made sense to me back then, and still make sense. So I'd like to understand further what differences you see between ion and dht.

PR #148, opened 2023-11-22
fix dht encoding

decentralgabe:
Small fix for dht coding to align with the spec. This will be tested in a follow-up that includes test vectors.


andresuribe87 (approved):
Would it make sense to add a test that failed, and now works?

decentralgabe:
@andresuribe87 prefer to do that when I do the test vectors...yes I imagine we will need that.

mistermoe (approved):
🏅

Issue #112, opened 2023-11-02
Dids - expose static `load` and `create` methods instead of loading via constructor and constructing net new via `.create`


phoebe-lew:

phoebe-lew:
This was due to a bunch of confusion on BTC100 side when using the SDKs - they thought that calling the constructor would give them a net new DID, and `create` with existing key material passed in would effectively "load" and existing one. 

The suggestion is to make `load` more explicit than just using the constructor. `create` can probably stay the same.

andresuribe87:
I see your point. That's definitely a different kind of load than the one I understood.

An orthogonal consideration, which I noticed while doing the PR, is that we currently have no way of validating whether the `keyManager` contains the necessary material to be able to effectively use it as the backer of the Did. That's something that I thought was very important, and made the client's life's easier.

mistermoe:
agreed on validation @andresuribe87 . might as well validate asap vs. waiting for things to bork bc the right keys aren't present when `sign` is called at some point

andresuribe87:
Summarizing the different alternatives here. With some changes in naming, we could actually provide both pieces of functionality.

## Alternatives

### `importAndLoad` accepts a `keySet`

This is what is being proposed in comment https://github.com/TBD54566975/web5-kt/issues/112#issuecomment-1819870135. One implication is that the `InMemoryKeyManager.import` method would need to be elevated to be part of `KeyManager`, and we'll have to force all implementations to have an implementation of it. Not a big deal, but I'm always concerned when expanding the public API surface, specially of interfaces.

The implementation is spelled out in in the same comment. We would additionally need to introduce validation in the constructor (or somewhere else).

The advantage of this is that it does all the things that consumers indicated they were doing.

Downsides include:
* Passing a full  `keySet` to load a single `DidKey` strikes me as not being intuitive. 
* All methods that return a `Did` currently accept a `keyManager`. This would be the first time we're introducing something different. It might harm dev-ex. 

### `load` accepts a `keyManager`

This is what was proposed in #139.

The advantage of this approach is consistency. Additionally, I think doing non-trivial validation logic is something that I wouldn't expect a constructor to do. Though it's fine as long as it's well documented, and nice errors are returned.

Clear disadvantage is that it doesn't fully solve the problem described.

## Validation

In either implementation, we agree that it's important to validate asap before consumers try to sign.

## Decisions to make

* Do we implement both mechanisms?
* Do we keep constructors public and do validation inside them?

[...]

PR #152, opened 2023-11-30
Allow public constructors for Did subclasses

andresuribe87:
While we come to consensus on #112 , we're allowing constructors to be public of subclasses of `Did` to be public.


KendallWeihe (approved):
thanks!

amika-sq (approved):

PR #153, opened 2023-12-02
Added tests in the style of test-vectors for credential creation

andresuribe87:
# Overview
This is in preparation before promoting them to the `sdk-development` repo.

# Description
This is part of our push to have test vectors. As a result of adding them, I had to add validation to the `fromJson` function. This is the first step in the direction to have test vectors for verifiable credentials 1.1. Heavily influenced from https://github.com/w3c/vc-test-suite/blob/gh-pages/test/vc-data-model-1.0/input/example-017-missing-verifiableCredential.jsonld

# How Has This Been Tested?
This is simply adding more testing area.


credentials/src/main/kotlin/web5/sdk/credentials/VerifiableCredential.kt
```diff
@@ -306,7 +309,108 @@ public class VerifiableCredential internal constructor(public val vcDataModel: V
...
      */
     public fun fromJson(vcJson: String): VerifiableCredential {
-      return VerifiableCredential(VcDataModel.fromJson(vcJson))
+      val typeRef = object : TypeReference<HashMap<String, Any>>() {}
+      val vcMap = objectMapper.readValue(vcJson, typeRef)
+      validateDataModel(vcMap)
+      return VerifiableCredential(VcDataModel.fromMap(vcMap))
+    }
+
+    private fun validateDataModel(model: Map<String, Any>) {
```
amika-sq, line 318: Should we be using the `@Throws` annotation here? Seems like there's two possible outcomes for this function:
1. The function completes
2. The function throws

That didn't really occur to me on my first read-through that it'd throw. Would be great to add for clarity for readers!

credentials/src/test/kotlin/web5/sdk/credentials/StatusListCredentialTest.kt
```diff
@@ -27,17 +27,8 @@ class StatusListCredentialTest {
...
     )
 
-    assertTrue(
-      specExampleRevocableVc.vcDataModel.contexts.containsAll(
-        listOf(
-          URI.create("https://www.w3.org/ns/credentials/v2"),
-          URI.create("https://www.w3.org/ns/credentials/examples/v2")
-        )
-      )
-    )
```
amika-sq, line 37: Should this be removed? Seems unrelated to the test vectors on first glance.

credentials/src/test/kotlin/web5/sdk/credentials/VerifiableCredentialTest.kt
```diff
@@ -94,40 +101,75 @@ class VerifiableCredentialTest {
...
+      """{
+          "kty": "EC",
+          "d": "OeBCt3M8roz9F9Ny192TtZXf_-qPzAkwSM6ep7QdSO4",
+          "use": "sig",
+          "crv": "secp256k1",
+          "kid": "NB_qlUIr06-AikVVMFreq0lc-omQtzc6lwhhcvgO6r4",
+          "x": "DdtN8W6x_34pB_nkxR0e1tmDkNnsJeusBAEPzKWgf_Y",
+          "y": "u3W135inodLqtcEb9jNGS3JsM_uFKmkJSb8Trc9luWI",
+          "alg": "ES256K"
+        }"""
```
amika-sq, line 115: I'm not seeing this being used anywhere. Still needed?

credentials/src/test/kotlin/web5/sdk/credentials/VerifiableCredentialTest.kt
```diff
@@ -94,40 +101,75 @@ class VerifiableCredentialTest {
...
+      assertFails(vector.description) {
+        VerifiableCredential.fromJson(mapper.writeValueAsString(testInput.credential))
+      }
+    }
+  }
+
+  @Test
+  fun foo() {
+    assertFalse(URI("example").isAbsolute)
   }
```
amika-sq, line 173: Still needed?

credentials/src/test/kotlin/web5/sdk/credentials/VerifiableCredentialTest.kt
```diff
@@ -94,40 +101,75 @@ class VerifiableCredentialTest {
...
+    }
 
-    assertEquals(holderDid.uri, vcDataModel.credentialSubject.id.toString())
-    assertContains(vcDataModel.types, "StreetCred")
+    testVectors.vectors.filter { it.errors ?: false }.forEach { vector ->
+      val testInput = vector.input
+      assertFails(vector.description) {
+        VerifiableCredential.fromJson(mapper.writeValueAsString(testInput.credential))
+      }
+    }
```
amika-sq, line 167: Just a thought on structure here:

We could have two test vectors: `create-success.json` and `create-failure.json`. That way we could have separate functions that test success/failure scenario scenarios instead of filtering on the inputs. 

But I'm ok with this too! I'm sure our test vectors will evolve over time, so definitely not blocking by any means.

dids/src/main/kotlin/web5/sdk/dids/extensions/DidExtensions.kt
```diff
@@ -0,0 +1,27 @@
...
+import web5.sdk.dids.methods.key.DidKey
+import web5.sdk.dids.methods.web.DidWeb
+
+internal val supportedMethods = mapOf(
+  DidKey.methodName to DidKey.Companion,
+  DidJwk.methodName to DidJwk.Companion,
+  DidIon.methodName to DidIon.Default,
+  DidDht.methodName to DidDht.Default,
+  DidWeb.methodName to DidWeb.Default
+)
```
amika-sq, line 18: This looks like the same values as `methodResolves` in the `DidResolvers.kt` class. Wondering if it might make sense to combine them?

amika-sq (approved):
Some comments, but nothing blocking from me. Looking forward to having more test vectors!

credentials/src/test/kotlin/web5/sdk/credentials/VerifiableCredentialTest.kt
```diff
@@ -94,40 +101,75 @@ class VerifiableCredentialTest {
...
+  data class TestInput(
+    val signerDidUri: String?,
+    val signerPrivateJwk: Map<String, Any>?,
+    val credential: Map<String, Any>?,
+  )
 
-    val vcDataModelJson = jwtClaims.getJSONObjectClaim("vc")
-    val vcDataModel = VcDataModel.fromMap(vcDataModelJson)
+  @Test
+  fun `creates test vectors`() {
```
KendallWeihe, line 143: Slightly misleading name, in that, it could be interpreted that this function creates the test vectors JSON

Maybe...

```suggestion
  fun `test vectors: creates`() {
```

KendallWeihe:
Is our `validateDataModel` based on a spec? Or is it a TBD prescription? 

If it's an encoding of a spec, could we link to the spec?

credentials/src/test/kotlin/web5/sdk/credentials/StatusListCredentialTest.kt
andresuribe87: I removed it because we were creating a VC that was following the 2.0 model, but we're implementing the 1.1 version of VCs. 

It goes hand in hand with the changes to the `revocable_vc.json` file.

credentials/src/main/kotlin/web5/sdk/credentials/VerifiableCredential.kt
andresuribe87: Excellent point, added comments. Let me know if it's more clear now.

credentials/src/test/kotlin/web5/sdk/credentials/VerifiableCredentialTest.kt
andresuribe87: It wasn't, thanks for catching! Removed

credentials/src/test/kotlin/web5/sdk/credentials/VerifiableCredentialTest.kt
andresuribe87: Removed.

credentials/src/test/kotlin/web5/sdk/credentials/VerifiableCredentialTest.kt
andresuribe87: Unforturnately, cannot include `:` chars in the test names. Instead, I went with the approach of defining a new class as shown below. Open to suggestions/feedback!

```kt
class TestVectorsCredentialsTest {
  @Test
  fun creates() {
// ...
  }
}
```

credentials/src/test/kotlin/web5/sdk/credentials/VerifiableCredentialTest.kt
andresuribe87: I like that! Used `_` instead of `-` because it's what was allowed :)

dids/src/main/kotlin/web5/sdk/dids/extensions/DidExtensions.kt
andresuribe87: Combined!

andresuribe87:
> Is our `validateDataModel` based on a spec? Or is it a TBD prescription?
> 
> If it's an encoding of a spec, could we link to the spec?

@KendallWeihe I added a comment clarifying and pointed to the spec. Let me know if that's clear now.

credentials/src/main/kotlin/web5/sdk/credentials/VerifiableCredential.kt
nitro-neal: This could live in a new validators file

andresuribe87: I would rather keep it private unless it's necessary to make this function public. Do you have a use-case in mind that requires it to be public?

Issue #144, opened 2023-11-20
Decide on approach for exceptions thrown by public methods


mistermoe:
We should decide on a convention we want to follow for exceptions thrown by public methods. 

Proposal:
* avoid explicitly throwing `Exception` and `RuntimeException`. Both are far too general, we should throw exceptions that indicate a specific error case. We're using `Exception` and `RuntimeException` in a handful of places at the moment. [example](https://github.com/TBD54566975/web5-kt/blob/c9bfb61ace04cb6a8f6afcbda6d148820c63e4ad/credentials/src/main/kotlin/web5/sdk/credentials/StatusListCredential.kt#L232)
* favor `Exception` over `Error` for custom exception classes. Conventionally, `*Error` are not expected to be caught. They are conditions that typically affect the JVM's ability to continue execution. Examples include `OutOfMemoryError` and `StackOverflowError`. `Error` is intended to indicate a serious problem. Example of us using the word `Error` [here](https://github.com/TBD54566975/web5-kt/blob/a644e6f129cabc85906099d8e79497a1d18fff71/credentials/src/main/kotlin/web5/sdk/credentials/PresentationExchange.kt#L74)
* ensure that exceptions thrown from our public api surface are either included as part of java/kotlin _or_ are custom exception classes that come directly from our package as opposed to third party dependencies

We should decide on our approach here and then prioritize implementing the approach before our codebase grows larger.

andresuribe87:
Super important topic, thanks for raising @mistermoe !

I agree with the first two points. 

Re: to 

> ensure that exceptions thrown from our public api surface are either included as part of java/kotlin or are custom exception classes that come directly from our package as opposed to third party dependencies

How do you do this in practice? For instance, if a dependency throws an custom exception, how can we guarantee that we'll not throw it? We would presumably need to surround most external dependency calls with `try/catch` blocks, but that's difficult to maintains because: (a) makes code incredibly verbose and (b) it's hard to enforce in PRs.

mistermoe:
great point @andresuribe87 . i was feeling 😅 🙈  as i typed that one out. 

Typically, what i've done is look at the javadoc/kdoc for method i'm calling to see if the exceptions thrown are documented. if they aren't documented, i click into the implementation to see if custom exceptions are thrown, but this can easily become a blackhole if the method i'm inspecting calls n other methods from other third party dependencies.

you're right, it does seem hard if not impossible to enforce in PRs. The motivation behind that bullet point is primarily two reasons:
[...]

PR #154, opened 2023-12-02
Custom exceptions instead of Exception, RuntimeException, and Error

jiyoontbd:
# Overview
Wrote custom exceptions and used them in place of throwing Exception, RuntimeException, and Error. [Linked issue](https://github.com/TBD54566975/web5-kt/issues/144)

# Description
Throwing typed exceptions with detailed error message, and annotating them in kdoc helps the consumer of the library understand why an exception was thrown and how to avoid the exception.

# How Has This Been Tested?
Wrote tests to ensure that new typed exceptions are thrown where expected.

# Checklist

- [x] I have read the CONTRIBUTING document.
- [x] My code is consistent with the rest of the project 
- [x] I have tagged the relevant reviewers and/or interested parties
- [x] I have updated the READMEs and other documentation of affected packages

## References
Updated kdocs to include `@throws` where necessary.

### DIFF
--- a/credentials/src/main/kotlin/web5/sdk/credentials/PresentationExchange.kt
+++ b/credentials/src/main/kotlin/web5/sdk/credentials/PresentationExchange.kt
@@ -43,8 +43,10 @@ public object PresentationExchange {
    * @param vcJwts Iterable of VCs in JWT format to validate.
    * @param presentationDefinition The Presentation Definition V2 object against which VCs are validated.
    * @throws UnsupportedOperationException If Submission Requirements are present in the definition.
-   * @throws PresentationExchangeException If the number of input descriptors matched is less than required.
+   * @throws PresentationExchangeException If the number of input descriptors matched is less than required
+   * or if the VC payload cannot be parsed as JSON.
    */
+  @Throws(UnsupportedOperationException::class, PresentationExchangeException::class)
   public fun satisfiesPresentationDefinition(
     vcJwts: Iterable<String>,
     presentationDefinition: PresentationDefinitionV2
--- a/credentials/src/main/kotlin/web5/sdk/credentials/PresentationExchange.kt
+++ b/credentials/src/main/kotlin/web5/sdk/credentials/PresentationExchange.kt
@@ -38,7 +38,7 @@ public object PresentationExchange {
    * This function ensures that the provided VCs meet the criteria defined in the Presentation Definition.
    * It first checks for the presence of Submission Requirements in the definition and throws an exception if they exist,
    * as this feature is not implemented. Then, it maps the input descriptors in the presentation definition to the
-   * corresponding VCs. If the number of mapped descriptors does not match the required count, an error is thrown.
+   * corresponding VCs. If the number of mapped descriptors does not match the required count, an exception is thrown.
    *
    * @param vcJwts Iterable of VCs in JWT format to validate.
    * @param presentationDefinition The Presentation Definition V2 object against which VCs are validated.
--- a/credentials/src/main/kotlin/web5/sdk/credentials/PresentationExchange.kt
+++ b/credentials/src/main/kotlin/web5/sdk/credentials/PresentationExchange.kt
@@ -43,7 +43,7 @@ public object PresentationExchange {
    * @param vcJwts Iterable of VCs in JWT format to validate.
    * @param presentationDefinition The Presentation Definition V2 object against which VCs are validated.
    * @throws UnsupportedOperationException If Submission Requirements are present in the definition.
-   * @throws PresentationExchangeError If the number of input descriptors matched is less than required.
+   * @throws PresentationExchangeException If the number of input descriptors matched is less than required.
    */
   public fun satisfiesPresentationDefinition(
     vcJwts: Iterable<String>,
@@ -90,7 +90,7 @@ public object PresentationExchange {
    * @param vcJwt The JWT string representing the Verifiable Credential.
    * @param inputDescriptor An instance of InputDescriptorV2 defining the criteria to be satisfied by the VC.
    * @return Boolean indicating whether the VC satisfies the criteria of the Input Descriptor.
-   * @throws PresentationExchangeError Any errors during processing
+   * @throws PresentationExchangeException If the VC payload cannot be parsed as JSON.
    */
   private fun vcSatisfiesInputDescriptor(
     vcJwt: String,
--- a/credentials/src/main/kotlin/web5/sdk/credentials/PresentationExchange.kt
+++ b/credentials/src/main/kotlin/web5/sdk/credentials/PresentationExchange.kt
@@ -8,7 +8,6 @@ import com.nfeld.jsonpathkt.JsonPath
 import com.nfeld.jsonpathkt.extension.read
 import com.nimbusds.jwt.JWTParser
 import com.nimbusds.jwt.SignedJWT
-import web5.sdk.credentials.exceptions.PresentationExchangeException
 
 /**
  * The `PresentationExchange` object provides functions for working with Verifiable Credentials
@@ -23,6 +22,7 @@ public object PresentationExchange {
    * @return A list of Verifiable Credentials that satisfy the Presentation Definition.
    * @throws UnsupportedOperationException If the method is untested and not recommended for use.
    */
+  @Throws(UnsupportedOperationException::class)
   public fun selectCredentials(
     credentials: List<VerifiableCredential>,
     presentationDefinition: PresentationDefinitionV2
@@ -94,6 +94,7 @@ public object PresentationExchange {
    * @return Boolean indicating whether the VC satisfies the criteria of the Input Descriptor.
    * @throws PresentationExchangeException If the VC payload cannot be parsed as JSON.
    */
+  @Throws(PresentationExchangeException::class)
   private fun vcSatisfiesInputDescriptor(
     vcJwt: String,
     inputDescriptor: InputDescriptorV2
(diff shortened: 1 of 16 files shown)


dids/src/main/kotlin/web5/sdk/dids/methods/web/DidWeb.kt
```diff
@@ -142,6 +143,6 @@ public sealed class DidWebApi(
   }
 
   public override fun create(keyManager: KeyManager, options: CreateDidOptions?): DidWeb {
-    throw RuntimeException("create operation not supported for did:web")
+    throw NotImplementedException("create operation not supported for did:web")
```
jiyoontbd, line 146: not sure what's a better exception to throw here.

dids/src/main/kotlin/web5/sdk/dids/methods/dht/DhtClient.kt
```diff
@@ -86,7 +88,8 @@ internal class DhtClient(
 
     val response = runBlocking { client.get("${gateway}/${id}") }
     if (!response.status.isSuccess()) {
-      throw RuntimeException("Error getting message from DHT")
+      val err = runBlocking { response.bodyAsText() }
+      throw PkarrRecordResponseException("Error reading Pkarr Record Set of id $id. Error: $err")
```
jiyoontbd, line 92: would love to know what id i should pass in when writing a unit test to check this exception is thrown. to do this, i need to pass in an `id` that's z-based 32 encoded bytes that when passed to diddht.tbddev.org/${id}, will return a 400, 404, or 500. so far i cannot get it to give me anything other than 200 🤔

credentials/src/main/kotlin/web5/sdk/credentials/PresentationExchange.kt
```diff
@@ -98,7 +101,7 @@ public object PresentationExchange {
     val vc = JWTParser.parse(vcJwt) as SignedJWT
 
     val vcPayloadJson = JsonPath.parse(vc.payload.toString())
-      ?: throw PresentationExchangeError("Failed to parse VC payload as JSON.")
+      ?: throw PresentationExchangeException("Failed to parse VC payload as JSON.")
```
jiyoontbd, line 104: would love to know how to write a test that will generate a wonky vc payload that can't be parsed correctly and throws this new exception

dids/src/main/kotlin/web5/sdk/dids/methods/dht/DhtClient.kt
mistermoe: cc: @decentralgabe

dids/src/main/kotlin/web5/sdk/dids/methods/web/DidWeb.kt
mistermoe: i think `NotImplementedException` makes a lot of sense!

nitro-neal (approved):

credentials/src/main/kotlin/web5/sdk/credentials/PresentationExchange.kt
jiyoontbd: ah right. i just used the example jwt in jwt.io but got rid of the payload section and that seemed to do the trick (since this line is trying to parse the vc.payload)

jiyoontbd:
> This is awesome! Just note that there was talk for the PresentationExchange Exceptions to be wrapped up in a `Results` object. I think exceptions may be better than the results object and easier for the end user.

ah, thanks for that context. that specific method `satisfiesPresentationDefinition` currently seems to be used as a `if i don't throw an exception, assume the the PD is satisfied`. this is ok with me, as long as we thoroughly explain why an exception was thrown and at what step.

personally, i feel just okay about this way of exception handling:

```
const resultObject = satisfiesPresentationDefinition(vcJwts, presentationDefinition)
if(resultObject.errors || resultObject.warnings) {
  throw new error()
}
```

i've tried to implement something like this and it just ended up being a bit duplicative and verbose at the end. having said that, there's a [great article](https://martinfowler.com/articles/replaceThrowWithNotification.html) that advocates for this approach, so i'm definitely open to future dev work that handles these exceptions differently :)

dids/src/main/kotlin/web5/sdk/dids/methods/dht/DhtClient.kt
jiyoontbd: followed up with gabe, seems like there's no easy way for us to come up with an ID that will pass the ZBase32.decode() but fail the client call to diddht.tbddev.org/{id}.

dids/src/main/kotlin/web5/sdk/dids/methods/web/DidWeb.kt
jiyoontbd: changed to unsupportedoperationexception :)

credentials/src/main/kotlin/web5/sdk/credentials/StatusListCredential.kt
```diff
@@ -68,7 +73,7 @@ public object StatusListCredential {
    * val statusListCredential = StatusListCredential.create("http://example.com/statuslistcred/id123", "http://example.com/issuers/1", StatusPurpose.REVOCATION, listOf(vc1,vc2))
    * ```
    */
-  @Throws(RuntimeException::class)
+  @Throws(StatusListCredentialCreateException::class)
```
KendallWeihe, line 76: Does the `@Throws` annotation automatically add info to the docs? Or do we need to add a `* @throws` comment in the lines above?

andresuribe87: It doesn't appear that it does. From https://tbd54566975.github.io/web5-kt/docs/htmlMultiModule/credentials/web5.sdk.credentials/-status-list-credential/create.html I see what's shown below.

image

credentials/src/main/kotlin/web5/sdk/credentials/exceptions/BitstringExpansionException.kt
andresuribe87, line 1: Not a blocker, but wanted to share that it should be perfectly acceptable to have multiple declarations per file. See  https://kotlinlang.org/docs/coding-conventions.html#source-file-names

If you're compelled, I would suggest putting all exceptions inside a `ExceptionDeclarations.kt` file.

credentials/src/test/kotlin/web5/sdk/credentials/PresentationExchangeTest.kt
```diff
@@ -234,9 +255,14 @@ class PresentationExchangeTest {
       )
       val vcJwt = vc.sign(issuerDid)
 
+      assertThrows<PresentationExchangeException> {
+        PresentationExchange.satisfiesPresentationDefinition(listOf(vcJwt), pd)
+      }
```
andresuribe87, line 260: Curious how this compares to the test below that calls `assertFailure`? Is the difference that the type is asserted vs the message?

jiyoontbd: agh. yea, i don't love the way i wrote this! really what i wanted was something like this, but i couldn't quite get it to work. so i'm wanting to test two different things - 1) that a specific exception was thrown, and 2) that it fails with a specific message

```
assertFailsWith< PresentationExchangeException> {
        PresentationExchange.satisfiesPresentationDefinition(listOf(vcJwt), pd)
      }.messageContains("Failed to parse VC payload as JSON.")
```

credentials/src/main/kotlin/web5/sdk/credentials/StatusListCredential.kt
jiyoontbd: added `@throws` in ktdocs and `@Throws` annotations

andresuribe87: Great candidate for another linting rule. Added https://github.com/TBD54566975/tbd-detekt-rules/issues/7

credentials/src/main/kotlin/web5/sdk/credentials/StatusListCredential.kt
```diff
@@ -221,17 +231,24 @@ public object StatusListCredential {
...
+          response,
+          "Failed to retrieve VerifiableCredentialType from $url"
+        )
       }
-    } catch (e: ClientRequestException) {
-      throw Exception("Failed to fetch the status list credential due to a request error: ${e.message}", e)
     } catch (e: ResponseException) {
-      throw Exception("Failed to fetch the status list credential due to a response error: ${e.message}", e)
+      // ClientRequestException will be caught here since it is a subclass of ResponseException
+      throw StatusListCredentialFetchException("Failed to fetch the status list credential: ${e.message}", e)
```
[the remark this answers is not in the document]
jiyoontbd: i noticed that this change was failing some tests, and then realized that removing the message would hide the underlying error message that's thrown from an upstream exception. 

i think this change would omit important info that the caller of this function would want to know, so I pulled it out but then added it back in.

mistermoe (approved):
💯

credentials/src/main/kotlin/web5/sdk/credentials/StatusListCredential.kt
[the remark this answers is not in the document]
KendallWeihe: I agree on that point @jiyoontbd good catch

credentials/src/test/kotlin/web5/sdk/credentials/PresentationExchangeTest.kt
andresuribe87: Gotcha! Perhaps it's worth making our own test assert file which would include this `assertFailsWith` function initially

Issue #110, opened 2023-11-01
replace usage of `nimbusds` `Algorithm` and `Curve` with our own enums


mistermoe:
Currently, `Crypto`, `KeyManager`, and `DidKey` expect nimbusds `JWSAlgorithm` and `Curve` arguments for various methods. we should implement our own enums and use those instead. surfacing a dependency's enums as arguments to our libs causes confusion. Confusion == kaka DevEX

andresuribe87:
> surfacing a dependency's enums as arguments to our libs causes confusion

@mistermoe can you elaborate on what confusion this is causing?

The way I think about it is:
* `JWSAlgorithm` represents any entry in the [JSON Web Signature and Encryption Algorithms table](https://www.iana.org/assignments/jose/jose.xhtml#web-signature-encryption-algorithms).
* `Curve` represents any entry in the [JSON Web Key Elliptic Curve table](https://www.iana.org/assignments/jose/jose.xhtml#web-key-elliptic-curve).

mistermoe:
yep, 

* `JWSAlgorithm` is confusing from a naming perspective
* we're asking callers to pass in types from third party libs
   * walking away from `nimbusds` will lead to breaking changes for our public api surface
   * if `nimbusds` changes these types in subsequent versions, we can get stuck on older versions or sort of be at the mercy of their public api surface

bradleydwyer:
> we're asking callers to pass in types from third party libs
walking away from nimbusds will lead to breaking changes for our public api surface
if nimbusds changes these types in subsequent versions, we can get stuck on older versions or sort of be at the mercy of their public api surface

@andresuribe87 This is the core issue for me. I believe if we are to provide tools for decentralised identity that we should commit to providing the types necessary to support that. I can also see a lot of benefit in a single narrative/style for how we present our SDKs, of which types are a significant part.

decentralgabe:
@mistermoe can you provide an example of our own enum to illustrate how it'd simplify things?

tomdaffurn:
We're exposing other Nimbus objects too, such as `JWK`

andresuribe87:
> Thoughts on making our own?

Yes, I think we should make our own. In general, I think we should follow https://jlbp.dev/JLBP-2. In particular, "Avoid exposing types from your dependencies".

So this issue can be split into two categories of tasks:
[...]

PR #155, opened 2023-12-02
Replace externally defined enums `Curve` and `Algorithm`

andresuribe87:
# Overview
This PR addresses #110 by introducing our own enums. 

# Description

Note that there are some cases in which the `nimbusds` classes are used. This is intentional, because the tests are actually already using it to create JWKs or JWTs directly.  The public API of the introduced enums is details below. 

### Curve

```kt
/**
 * Returns a [Curve] object given it's [ianaName]. The [ianaName] is how the curve was registered in the IANA table
 * available at https://www.iana.org/assignments/jose/jose.xhtml#web-key-elliptic-curve. When there is no Curve for
 * the given [ianaName], null is returned.
 */
public fun parse(ianaName: String): Curve?
```

Usage example:
```kt
val curve = Curve.parse("secp256k1")
assertEquals(Curve.SECP256K1, curve)

val nullCurve = Curve.pase("magicCurve")
assertNull(nullCurve)
```

### Algorithm

```kt
/**
 * Converts an [Algorithm] from the nimbusds library to a [Algorithm].
 */
public fun com.nimbusds.jose.Algorithm?.toWeb5Algorithm(): Algorithm?
```

This is useful when consumers have a nimbusds type, like what would represent a jwt, and want to be able to call our libraries.

 Usage example:
```kt
// Assume publicKeyJwk, toVerifyBytes, signatureBytes are all given already
val jwt = JWTParser.parse(vcJwt) as SignedJWT
Crypto.verify(publicKeyJwk, toVerifyBytes, signatureBytes, jwt.header.algorithm.toWeb5Algorithm())
```

# How Has This Been Tested?
This is refactoring, so unit tests already covered the changes.

### DIFF
--- a/crypto/src/main/kotlin/web5/sdk/crypto/Algorithm.kt
+++ b/crypto/src/main/kotlin/web5/sdk/crypto/Algorithm.kt
@@ -3,14 +3,14 @@ package web5.sdk.crypto
 /**
  * The supported JSON Web Signature algorithms.
  */
-public enum class JWSAlgorithm {
+public enum class Algorithm {
   /** ECDSA using secp256k1 curve and SHA-256. */
   ES256K,
 
   /** Edwards-curve Digital Signature Algorithm. */
   EdDSA;
 
-  internal fun toJwsAlgorithm(): com.nimbusds.jose.JWSAlgorithm {
+  internal fun toNimbusdsJWSAlgorithm(): com.nimbusds.jose.JWSAlgorithm {
     return when (this) {
       ES256K -> com.nimbusds.jose.JWSAlgorithm.ES256K
       EdDSA -> com.nimbusds.jose.JWSAlgorithm.EdDSA
--- a/crypto/src/main/kotlin/web5/sdk/crypto/Algorithm.kt
+++ b/crypto/src/main/kotlin/web5/sdk/crypto/Algorithm.kt
@@ -17,3 +17,10 @@ public enum class Algorithm {
     }
   }
 }
+
+/**
+ * Converts an [Algorithm] from the nimbusds library to a [Algorithm].
+ */
+public fun com.nimbusds.jose.Algorithm?.toWeb5Algorithm(): Algorithm? {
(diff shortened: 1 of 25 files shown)


crypto/src/main/kotlin/web5/sdk/crypto/Algorithm.kt
```diff
@@ -0,0 +1,26 @@
...
+      ES256K -> com.nimbusds.jose.JWSAlgorithm.ES256K
+      EdDSA -> com.nimbusds.jose.JWSAlgorithm.EdDSA
+    }
+  }
+}
+
+/**
+ * Converts an [Algorithm] from the nimbusds library to a [Algorithm].
+ */
+public fun com.nimbusds.jose.Algorithm?.toWeb5Algorithm(): Algorithm? {
```
nitro-neal, line 24: Well the naming is confusing enough but maybe this should just be called 

toAlgorithm(). Kind of strange to have web5 in the name

nitro-neal: jwt.header.algorithm.toAlgorithm() would be more ugly probably haha

andresuribe87: Yeah I'm not thrilled about the name either :(

I think `toWeb5Algorithm` is the "less bad". Also, I think we're considering a different interface from the conversation in https://github.com/TBD54566975/web5-rs/issues/38

phoebe-lew (approved):

frankhinek:
We can either merge this PR as is and then immediately throw up another PR that addresses https://github.com/TBD54566975/web5-kt/issues/197

or modify this PR to use a single algorithm identifier as is being done in `web5-js` and `web5-go`.

mistermoe:
Agreed I'd rather modify this one or close it out in favor of a new one that uses the ids we collectively landed on

mistermoe:
We're also looking to move away from nimbusds

PR #160, opened 2023-12-07
Patches security vulnerabilities

andresuribe87:
# Overview
WIP

# Description
_Include context, motivation, brief description, and an impact of the change(s). List follow-up tasks here._

# How Has This Been Tested?
_Describe the tests that you ran to verify your changes. Provide instructions for verification._

- [ ] Test A (e.g. Test A - New test that does ... run in ...)
- [ ] Test B

# Checklist

- [ ] I have read the CONTRIBUTING document.
- [ ] My code is consistent with the rest of the project 
- [ ] I have tagged the relevant reviewers and/or interested parties
- [ ] I have updated the READMEs and other documentation of affected packages

## References
_Please list relevant documentation (e.g. tech specs, articles, follow up or related work) relevant to this change, and note if the documentation has been updated._

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -5,12 +5,14 @@ import org.jetbrains.dokka.gradle.DokkaTaskPartial
 import org.jetbrains.kotlin.gradle.dsl.JvmTarget
 import java.net.URL
 
+
+
 plugins {
   id("org.jetbrains.kotlin.jvm") version "1.9.0"
   id("java-library")
   id("io.gitlab.arturbosch.detekt") version "1.23.1"
   `maven-publish`
-  id("org.jetbrains.dokka") version "1.9.0"
+  id("org.jetbrains.dokka") version "latest.release"
   id("org.jetbrains.kotlinx.kover") version "0.7.3"
   idea
 }
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -5,8 +5,6 @@ import org.jetbrains.dokka.gradle.DokkaTaskPartial
 import org.jetbrains.kotlin.gradle.dsl.JvmTarget
 import java.net.URL
 
-
-
 plugins {
   id("org.jetbrains.kotlin.jvm") version "1.9.0"
   id("java-library")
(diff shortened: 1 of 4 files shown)


dids/src/main/kotlin/web5/sdk/dids/methods/dht/DhtClient.kt
```diff
@@ -208,6 +205,15 @@ internal class DhtClient(
...
 
+    private fun bencode(bs: ByteArray): ByteArray {
+      val out = ByteArrayOutputStream()
+      val l = bs.size.toString()
+      out.write(l.toByteArray(charset("UTF-8")))
+      out.write(58)
+      out.write(bs)
+      return out.toByteArray()
+    }
+
```
mistermoe, line 216: can we add context for what this method does? maybe https://en.wikipedia.org/wiki/Bencode and include a test?

andresuribe87: Added docs and tests.

mistermoe:
this is great. only thing i'm hesitant of right now is 

> Changing the target JVM version from 11 to 17. This was done because 17 has better LTS. It also allowed us to upgrade `com.danubetech:verifiable-credentials-java` to version 1.7.0 until we hear feedback from one downstream consumer that's currently still on java 11

phoebe-lew (commented):
Do these fixes still work if we stay on 11?

andresuribe87:
> Do these fixes still work if we stay on 11?

@phoebe-lew most, but not all. Specifically, using the danubetech library for VCs has a transitive dep that's borked. It's fixed in a newer release, but that version of the lib is compiled for java 17.

andresuribe87:
@phoebe-lew @mistermoe  what I propose is to merge this without the upgrade to java 17. This fixes a bunch of things. We can deal with that once we have a clear answer on whether it's required.

nitro-neal (approved):
Looks good, if we are worried about backwards compatibility issues on anything we can use our trusty repo to try and consume it

https://github.com/andresuribe87/downgrading-hell

phoebe-lew (approved):

PR #169, opened 2023-12-13
Create presentation from credentials test vector in kt

nitro-neal:
This implements the test vectors found here : 
https://github.com/TBD54566975/web5-spec/blob/main/test-vectors/presentation_exchange/create_presentation_from_credentials.json

```kotlin
class Web5TestVectorsPresentationExchangeTest {
    @Test
    fun create_presentation_from_credentials() { }
}
```


credentials/src/test/kotlin/web5/sdk/credentials/PresentationExchangeTest.kt
```diff
@@ -578,4 +579,39 @@ class Web5TestVectorsPresentationExchange {
...
+      val vectorOutputPresSubmission = vector.output!!.presentationSubmission
+
+      assertEquals(vectorOutputPresSubmission.definitionId, presSubmission.definitionId)
+      assertEquals(vectorOutputPresSubmission.descriptorMap.size, presSubmission.descriptorMap.size)
+
+      for (i in vectorOutputPresSubmission.descriptorMap.indices) {
+        assertEquals(vectorOutputPresSubmission.descriptorMap[i].id, presSubmission.descriptorMap[i].id)
+        assertEquals(vectorOutputPresSubmission.descriptorMap[i].format, presSubmission.descriptorMap[i].format)
+        assertEquals(vectorOutputPresSubmission.descriptorMap[i].path, presSubmission.descriptorMap[i].path)
+      }
```
andresuribe87, line 614: Would this work? 
```suggestion
      assertEquals(vectorOutputPresSubmission, presSubmission, vector.description)
```

andresuribe87 (approved):

credentials/src/test/kotlin/web5/sdk/credentials/PresentationExchangeTest.kt
nitro-neal: nope:
```
Expected :web5.sdk.credentials.model.PresentationSubmission@62517709
Actual   :web5.sdk.credentials.model.PresentationSubmission@1fcdf8c1
```

I bet there is a deep equals or something let me look a bit

andresuribe87: That likely means you could make it work by overriding the equals method.

nitro-neal: making it a data class would fix that otherwise have to do the override hashcode / equals stuff.

However this still wont work in this case because the presSubmission ids will be different (by design) 

creating a new presentation definition makes a different presSubmission id guid

andresuribe87: > making it a data class would fix that otherwise have to do the override hashcode / equals stuff.

Yes, that's correct! The problem is that [binary backwards compatibility is broken](https://kotlinlang.org/docs/jvm-api-guidelines-backward-compatibility.html#don-t-use-data-classes-in-an-api) when using dataclasses.

> However this still wont work in this case because the presSubmission ids will be different (by design)

Ahhh, excellent point. It does feel a bit strange to have a test vector that's not compared exactly, but I guess that's fine as long as it's documented in the README!

phoebe-lew (approved):

PR #170, opened 2023-12-13
JWT and JsonPath parsing done outside the loop

andresuribe87:
This is a trivial change so we do less work inside a loop.


phoebe-lew (approved):
🙏

nitro-neal (approved):

PR #177, opened 2023-12-19
Add PEX Validate Definition

nitro-neal:
# Overview
Added a validator to validate the presentation definition

# Description
Does many validations according to the spec - https://identity.foundation/presentation-exchange/#presentation-submission and our javascript library to have rules for validating a presentationDefinition


credentials/src/main/kotlin/web5/sdk/credentials/model/Validator.kt
```diff
@@ -0,0 +1,125 @@
...
+ **/
+public object PresentationDefinitionV2Validator {
+
+  /**
+   * Validates a PresentationDefinitionV2.
+   */
+  public fun validate(presentationDefinition: PresentationDefinitionV2) {
+    require(presentationDefinition.id.isNotEmpty()) { "PresentationDefinition id must not be empty" }
+    presentationDefinition.name?.let { require(it.isNotEmpty()) {
+      "PresentationDefinition name must not be empty" }
```
andresuribe87, line 19: The approach we've settled on is to use typed exceptions for this sort of stuff. See https://github.com/TBD54566975/web5-kt/blob/48ce6764163b32714be7d8282bfad0678f49debb/credentials/src/main/kotlin/web5/sdk/credentials/ExceptionDeclarations.kt#L10 for an example definition and use. 

Would it make sense to move towards having definitions for all exceptions thrown?

credentials/src/main/kotlin/web5/sdk/credentials/model/Validator.kt
```diff
@@ -0,0 +1,125 @@
...
+import web5.sdk.credentials.JsonPathParseException
+
+/**
+ * PresentationDefinitionV2Validator Validator.
+ **/
+public object PresentationDefinitionV2Validator {
+
+  /**
+   * Validates a PresentationDefinitionV2.
+   */
```
andresuribe87, line 15: Here and elsewhere, I'd suggest adding which validations are done. Either explicitly, or pointing to the spec.

credentials/src/test/kotlin/web5/sdk/credentials/PresentationDefinitionTest.kt
andresuribe87, line 1: Is the intention to add test vectors later?

phoebe-lew (commented):
would this be jayyyson-schema-able?

nitro-neal:
Yea at first I was going for the json schema approach, but there are a lot of nuances which I don't think json schema can accomplish:

* All inputDescriptor ids must be unique
* Field ids must be unique across all input descriptors
* Checking the internal JSON path

and more

credentials/src/main/kotlin/web5/sdk/credentials/model/Validator.kt
nitro-neal: Yess will update 👍

credentials/src/test/kotlin/web5/sdk/credentials/PresentationDefinitionTest.kt
nitro-neal: Added 👍

credentials/src/main/kotlin/web5/sdk/credentials/ExceptionDeclarations.kt
```diff
@@ -41,3 +41,11 @@ public class StatusListCredentialFetchException(cause: Throwable, message: Strin
  */
 public class StatusListCredentialParseException(cause: Throwable, message: String? = null)
   : RuntimeException(message, cause)
+
+/**
+ * Presentation exchange validation exception.
+ *
+ * @param message the exception message detailing the error
+ */
+public class PexValidationException(message: String? = null)
```
andresuribe87, line 50: Thought on having the type of the exception be specific to the error? For example:

```kt
public class PresentationDefinitionIdEmptyException
```

I see that being the approach taken for the existing exceptions. One benefit is that clients can handle those exceptions by type programatically, instead of having to parse the message.

andresuribe87 (approved):
Good stuff. 

One comment about exceptions, otherwise lgtm.

credentials/src/main/kotlin/web5/sdk/credentials/ExceptionDeclarations.kt
nitro-neal: yes I was thinking about this, but then we would have like a new exception for each different check basically. 15 different Exception classes seems excessive right?

andresuribe87: I'd say it depends. For example, did resolution has 10 different errors https://w3c-ccg.github.io/did-resolution/#errors

Perhaps some could be combined, so there's a balance in the middle. Up to you.

phoebe-lew (approved):

PR #181, opened 2023-12-28
Did dht test vectors

andresuribe87:
# Overview
Adds test vectors for did:dht for the `create` and `resolve` functions. The motivation is to ensure interoperability across implementations of the standard. This PR is related to https://github.com/TBD54566975/sdk-development/issues/11, and will form the basis to fix the issue.

# Description
This PR includes the following changes:
* Adds typed exceptions to `ExceptionDeclarations.kt`.
* Adds test vector files for `create` and `resolve`.
* Modifies the implementation of `DidDht` so that tests pass.
* Adds a README for the test-vectors added.

Once this is accepted, the test vectors will be promoted to the `sdk-development` repository.

> [!NOTE]
> The test vectors were based off of https://did-dht.com/#test-vectors, and adjusted to match our test-vector structure.

# How Has This Been Tested?

Unit tests


dids/src/main/kotlin/web5/sdk/dids/methods/dht/DidDht.kt
```diff
@@ -206,12 +212,32 @@ public sealed class DidDhtApi(configuration: DidDhtConfiguration) : DidMethod<Di
...
    * @return A [DidResolutionResult] instance containing the DID Document and related context, including types
    * as part of the [DidDocumentMetadata], if available.
-   * @throws IllegalArgumentException if the provided DID does not conform to the "did:dht" method.
    */
   override fun resolve(did: String, options: ResolveDidOptions?): DidResolutionResult {
-    validate(did)
+    return try {
+      resolveInternal(did)
+    } catch (e: Exception) {
+      println(e)
```
nitro-neal, line 220: remove

andresuribe87: Replaced with a logger.

test-vectors/did_dht/create.json
```diff
@@ -0,0 +1,150 @@
+{
+  "description": "did:dht resolution",
+  "vectors": [
+    {
+      "description": "resolves the vector in https://did-dht.com/#vector-1",
```
frankhinek, line 5: Shouldn't the vector description be about creation and not resolution?

andresuribe87: Good catch - I've updated this in [b119689](https://github.com/TBD54566975/web5-kt/pull/181/commits/b11968922fa62e9a9ac3aee4ea0c723f5005a0f1)

frankhinek:
I could be misreading the Kotlin (KT newb 🖐️)...

It seems that the `create` operation test vector provides an identity key as input:
https://github.com/TBD54566975/web5-kt/blob/02932bec4b15a9f278f87eb0210831789a188ab9/test-vectors/did_dht/create.json#L7-L13

But the `DidDht` implementation doesn't support specifying a pre-existing identity key and mocks the key generation so that the test vector passes:
https://github.com/TBD54566975/web5-kt/blob/02932bec4b15a9f278f87eb0210831789a188ab9/dids/src/test/kotlin/web5/sdk/dids/methods/dht/DidDhtTest.kt#L368-L369

Had a number of long discussions with Moe and Adam this week on how to approach DID key management across all of our SDKs.  It seems there's an opportunity to continue to get feedback from our partners and tweak how `create()`, `load()`, etc. work.  But that should probably wait for another day and not block this PR.

andresuribe87:
> I could be misreading the Kotlin (KT newb 🖐️)...
> 
> It seems that the `create` operation test vector provides an identity key as input:
> 
> https://github.com/TBD54566975/web5-kt/blob/02932bec4b15a9f278f87eb0210831789a188ab9/test-vectors/did_dht/create.json#L7-L13
> 
> But the `DidDht` implementation doesn't support specifying a pre-existing identity key and mocks the key generation so that the test vector passes:
> 
> https://github.com/TBD54566975/web5-kt/blob/02932bec4b15a9f278f87eb0210831789a188ab9/dids/src/test/kotlin/web5/sdk/dids/methods/dht/DidDhtTest.kt#L368-L369
> 
> Had a number of long discussions with Moe and Adam this week on how to approach DID key management across all of our SDKs. It seems there's an opportunity to continue to get feedback from our partners and tweak how `create()`, `load()`, etc. work. But that should probably wait for another day and not block this PR.

Your reading is correct. The dht identity key cannot be passed in. 

There were also other long discussions about the same topic during the ION implementation. The decision at that point in time is that the sdk should allow both paradigms: manually specifying the keys and also automatically creating them. That's the design that motivated https://github.com/TBD54566975/web5-kt/blob/e822caf27dc735be8c76edbac39a452046aecea2/dids/src/main/kotlin/web5/sdk/dids/methods/ion/DidIon.kt#L794

I agree that such a discussion shouldn't block this.

phoebe-lew (approved):
@frankhinek might be good for us to schedule time to discuss key management across the SDK?

PR #192, opened 2024-01-19
Add submodules

nitro-neal:
Create submodule for test vectors pointing to https://github.com/TBD54566975/web5-spec

Similar to this but on the web5-kt side - https://github.com/TBD54566975/web5-js/pull/381

Once this looks good I will update the readme


andresuribe87 (approved):

andresuribe87:
@nitro-neal what do you think about adding the git clone instructions to the README, so that users know they need to recursively checkout submodules?

jiyoontbd (approved):

PR #208, opened 2024-01-25
Validate presentaion submission with new validator

nitro-neal:
This method performs several checks to ensure the integrity of the presentation submission model object:
   * 1. Ensures that the presentation submission's ID is not empty.
   * 2. Validates that the definitionId is not empty.
   * 3. Validates descriptorMap is a non-empty list.
   * 4. Verifies the input descriptor mapping ids are the same on all levels of nesting.
   * 5. Ensures that the path is valid across all levels of nesting
   
 Usage:
 ```
 try {
   PresentationExchange.validateSubmission(presentationSubmission)
 } catch(Exception e) {
   // invalid submission
 }
 ```

web5-spec vector change - https://github.com/TBD54566975/web5-spec/pull/109

### DIFF
--- a/credentials/src/main/kotlin/web5/sdk/credentials/PresentationExchange.kt
+++ b/credentials/src/main/kotlin/web5/sdk/credentials/PresentationExchange.kt
@@ -146,8 +146,9 @@ public object PresentationExchange {
    * 1. Ensures that the presentation submission's id is not empty.
    * 2. Validates that the definitionId is not empty.
    * 3. Validates descriptorMap is a non-empty list.
-   * 4. Verifies the input descriptor mapping ids are the same on all levels of nesting.
-   * 5. Ensures that the path is valid across all levels of nesting
+   * 4. Check for unique inputDescriptor ids at top level
+   * 5. Verifies the input descriptor mapping ids are the same on all levels of nesting.
+   * 6. Ensures that the path is valid across all levels of nesting
    *
    * Throws an [PexValidationException] if the provided object does not conform to the Presentation Definition
    */
--- a/credentials/src/main/kotlin/web5/sdk/credentials/PresentationExchange.kt
+++ b/credentials/src/main/kotlin/web5/sdk/credentials/PresentationExchange.kt
@@ -12,6 +12,7 @@ import web5.sdk.credentials.model.InputDescriptorV2
 import web5.sdk.credentials.model.PresentationDefinitionV2
 import web5.sdk.credentials.model.PresentationDefinitionV2Validator
 import web5.sdk.credentials.model.PresentationSubmission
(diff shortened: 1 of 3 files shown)


credentials/src/main/kotlin/web5/sdk/credentials/model/Validator.kt
```diff
@@ -149,3 +149,70 @@ public object FieldV2Validator {
...
+    if (descriptor.path.isEmpty()) {
+      throw PexValidationException("Descriptor path should not be empty")
+    }
+
+    if (descriptor.format.isEmpty()) {
+      throw PexValidationException("Descriptor format should not be empty")
+    }
+
+    if (descriptor.id != id) {
+      throw PexValidationException("Each descriptor should have one id in it, on all levels")
```
decentralgabe, line 205: also need to check that descriptor ids are unique within the definition

credentials/src/main/kotlin/web5/sdk/credentials/PresentationExchange.kt
```diff
@@ -136,6 +137,26 @@ public object PresentationExchange {
...
+   * 2. Validates that the definitionId is not empty.
+   * 3. Validates descriptorMap is a non-empty list.
+   * 4. Check for unique inputDescriptor ids at top level
+   * 5. Verifies the input descriptor mapping ids are the same on all levels of nesting.
+   * 6. Ensures that the path is valid across all levels of nesting
+   *
+   * Throws an [PexValidationException] if the provided object does not conform to the Presentation Definition
+   */
+  @Throws(PexValidationException::class)
+  public fun validateSubmission(presentationSubmission: PresentationSubmission) {
```
angiejones, line 156: this seems to differ from the JS version which returns a result vs throwing an exception

```js
const submissionCheck = PresentationExchange.validateSubmission({
    presentationSubmission: presentationResult.presentationSubmission
  });
  ```
  
  Any reason why?

nitro-neal: So actually the initial thought on all the verifies was to throw an exception if it failed, and not have the results object.

In JS I happened to use the raw underlying Results object instead of wrapping it with an exception (maybe a mistake)

To keep it consistent I thinik we should decide if we want to use a `Results` object everywhere, OR if we want to 'fix' the JS and remove the existing results object and go back to throwing an exception

Tracking the discussion on that decision here - https://github.com/TBD54566975/web5-js/issues/386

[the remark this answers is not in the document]
nitro-neal: Nope there is only one version of submission and honestly we should probably remove the v2 from definition since I think we will only be using v2

PR #213, opened 2024-02-02
Adding tests for Ed25519 vectors

kirahsapong:
# Overview
Adds tests for Ed25519 `sign` and `verify` vectors and extends `validateKey` into two new functions `validatePrivateKey` and `validatePublicKey`

# Description
Web5TestVectorsCryptoEd25519
sign()
verify()

Adds tests to SDK report runner
https://tbd54566975.github.io/sdk-report-runner/

### DIFF
--- a/crypto/src/main/kotlin/web5/sdk/crypto/Ed25519.kt
+++ b/crypto/src/main/kotlin/web5/sdk/crypto/Ed25519.kt
@@ -192,6 +192,5 @@ public object Ed25519 : KeyGenerator, Signer {
    */
   private fun validateKey(key: JWK) {
     require(key is OctetKeyPair) { "key must be an Octet Key Pair (kty: OKP)" }
-    require(key.keyType == keyType) { "private key key type must be OKP" }
   }
 }
--- a/crypto/src/test/kotlin/web5/sdk/crypto/Ed25519Test.kt
+++ b/crypto/src/test/kotlin/web5/sdk/crypto/Ed25519Test.kt
@@ -63,7 +63,6 @@ class Web5TestVectorsCryptoEd25519 {
     val testVectors = mapper.readValue(File("../web5-spec/test-vectors/crypto_ed25519/verify.json"), typeRef)
 
     testVectors.vectors.filter { it.errors == false }.forEach { vector ->
-      println("Testing vector for ${vector.description}")
       val key = OctetKeyPair.parse(vector.input.key)
       val data = hexStringToByteArray(vector.input.data)
       val signature = hexStringToByteArray(vector.input.signature)
@@ -80,7 +79,6 @@ class Web5TestVectorsCryptoEd25519 {
 
     testVectors.vectors.filter { it.errors == true }.forEach { vector ->
       assertFails {
-        println("Testing vector for ${vector.description}")
         val key = OctetKeyPair.parse(vector.input.key)
         val data = hexStringToByteArray(vector.input.data)
         val signature = hexStringToByteArray(vector.input.signature)


nitro-neal (approved):
Looks good to me.
Just fyi to other reviewers, this change makes it so that you cannot verify() with a private key anymore, (which doesn't really make sense)

This is inline with the spec and has the same functionally now as web5-js

crypto/src/main/kotlin/web5/sdk/crypto/Ed25519.kt
```diff
@@ -156,8 +190,8 @@ public object Ed25519 : KeyGenerator, Signer {
    * @param key The [JWK] to validate.
    * @throws IllegalArgumentException if the key is not of type [OctetKeyPair] or if the key type is not [KeyType.EC].
    */
-  public fun validateKey(key: JWK) {
-    require(key is OctetKeyPair) { "private key must be an Octet Key Pair (kty: OKP)" }
+  private fun validateKey(key: JWK) {
+    require(key is OctetKeyPair) { "key must be an Octet Key Pair (kty: OKP)" }
     require(key.keyType == keyType) { "private key key type must be OKP" }
```
frankhinek, line 195: Since the `validateKey()` function is shared by both `validatePrivateKey()` and `validatePublicKey()` should this error not reference "private" since it may be raised when checking whether a public key's `kty` is `OKP`?

```suggestion
    require(key.keyType == keyType) { "key type must be OKP" }
```

crypto/src/main/kotlin/web5/sdk/crypto/Ed25519.kt
```diff
@@ -156,8 +190,8 @@ public object Ed25519 : KeyGenerator, Signer {
    * @param key The [JWK] to validate.
    * @throws IllegalArgumentException if the key is not of type [OctetKeyPair] or if the key type is not [KeyType.EC].
    */
-  public fun validateKey(key: JWK) {
-    require(key is OctetKeyPair) { "private key must be an Octet Key Pair (kty: OKP)" }
+  private fun validateKey(key: JWK) {
+    require(key is OctetKeyPair) { "key must be an Octet Key Pair (kty: OKP)" }
```
frankhinek, line 194: I'm not as familiar with what is being done with these checks but are these two lines redundantly checking for the same thing? (i.e., that `kty` is OKP?)

Or does `key is OctetKeyPair` go beyond just checking the `kty` property?

```kotlin
    require(key is OctetKeyPair) { "key must be an Octet Key Pair (kty: OKP)" }
    require(key.keyType == keyType) { "private key key type must be OKP" }
```

phoebe-lew: I think the first one is a runtime type check. `OctetKeyPair` extends `JWK` so it's possible that `key` does not conform to OKP type. 

Not sure if we need the second check if the first check is true though.

jiyoontbd (approved):

crypto/src/test/kotlin/web5/sdk/crypto/Ed25519Test.kt
```diff
@@ -0,0 +1,107 @@
...
+    }
+  }
+
+  @Test
+  fun verify() {
+    val typeRef = object : TypeReference<TestVectors<VerifyTestInput, Boolean>>() {}
+    val testVectors = mapper.readValue(File("../web5-spec/test-vectors/crypto_ed25519/verify.json"), typeRef)
+
+    testVectors.vectors.filter { it.errors == false }.forEach { vector ->
+      println("Testing vector for ${vector.description}")
```
nitro-neal, line 66: remove

nitro-neal, line 66:
```suggestion

```

crypto/src/test/kotlin/web5/sdk/crypto/Ed25519Test.kt
```diff
@@ -0,0 +1,107 @@
...
+      } else {
+        assertFails {
+          Ed25519.verify(key, data, signature)
+        }
+      }
+    }
+
+    testVectors.vectors.filter { it.errors == true }.forEach { vector ->
+      assertFails {
+        println("Testing vector for ${vector.description}")
```
nitro-neal, line 83:
```suggestion

```

nitro-neal (approved):

PR #219, opened 2024-02-06
Extracts Publish Tests Report Action

leordev:
Same as https://github.com/TBD54566975/tbdex-kt/pull/150

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,42 +52,11 @@ jobs:
           verbose: true
           flags: ${{ runner.os }}
 
-      - name: Publish Tests Results
-        env:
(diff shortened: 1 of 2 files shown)


jiyoontbd (approved):

nitro-neal (approved):

nitro-neal:
oh I forgot to check, but they run one after another right? The artifact needs to be uplaoded before it kicks off

jiyoontbd:
@leordev is this good to merge? also wondering if neal's q above is answered 
> oh I forgot to check, but they run one after another right? The artifact needs to be uplaoded before it kicks off

leordev:
> oh I forgot to check, but they run one after another right? The artifact needs to be uplaoded before it kicks off

Yes, correct!

@jiyoontbd thanks for the ping, it was on the sea of my GH inbox notifications

Issue #226, opened 2024-02-10
Align versions across Web5 platform


ALRubinger:
Uplevel version declarations at the platform level, shared across all Web5 submodules. Export the dependencies in a BOM POM as part of the Maven publishing step so it's consumable by projects building atop Web5. 

Concretely speaking, this amounts to a `dependencyManagement` section in the generated Maven POM of the `web5` parent. Submodules will pick up versioning from the parent in both Gradle and Maven builds. 

Will issue a PR when done.

ALRubinger:
For now I'm going to descope this from a Maven BOM POM build output in favor of Gradle's Version Catalogs. This will take care of version centralization, aligning versions across the submodules in the platform. Incrementally can look at Maven BOM POM porting once that work is done - it's less urgent.

WIP is in the commits as noted on this issue, in `upstream` branch `issue-226/version-alignment`.

ALRubinger:
Before closing this issue, open up new issues for uncovered API leaks discussed with @mistermoe. Doc those API leaks in the build with references to those new issues so they can be easily handled when the team is ready. PR for this issue will contain those comments and links.

ALRubinger:
617bea95b9732e3fedc6efe800f7886fc7f51b92 via #227

PR #227, opened 2024-02-11
Version Alignment and Introducing Gradle Version Catalogs

ALRubinger:
Centralizes versions across the `web5-kt` submodules using Gradle Version Catalogs.

Kicking off draft PR now to get early authoritative CI feedback during development.

Addresses:
* #226

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -6,8 +6,8 @@ import org.jetbrains.kotlin.gradle.dsl.JvmTarget
 import java.net.URL
 
 plugins {
-  id("org.jetbrains.kotlin.jvm") version "1.9.+"
-  id("java-library")
+  id("org.jetbrains.kotlin.jvm") version "1.9.22"
+  id("base")
   id("io.gitlab.arturbosch.detekt") version "1.23.+"
   `maven-publish`
   id("org.jetbrains.dokka") version "1.9.+"
@@ -15,6 +15,7 @@ plugins {
   signing
   idea
   id("io.github.gradle-nexus.publish-plugin") version "1.3.0"
+  id("version-catalog")
 }
 
 configurations.all {
@@ -150,7 +151,7 @@ subprojects {
       create<MavenPublication>(publicationName) {
         groupId = project.group.toString()
         artifactId = name
-        description = "Kotlin SDK for web5 functionality"
+        description = name
         version = project.property("version").toString()
         from(components["java"])
       }
@@ -158,7 +159,7 @@ subprojects {
         pom {
           name = publicationName
           packaging = "jar"
-          description.set("web5 kotlin SDK")
+          description.set("Web5 SDK for the JVM")
           url.set("https://github.com/TBD54566975/web5-kt")
           inceptionYear.set("2023")
           licenses {
@@ -240,14 +241,14 @@ publishing {
     create<MavenPublication>("web5") {
       groupId = project.group.toString()
       artifactId = name
-      description = "Kotlin SDK for web5 functionality"
+      description = "Web5 SDK for the JVM"
       version = project.property("version").toString()
       from(components["java"])
 
       pom {
-        packaging = "jar"
-        name = project.name
-        description.set("web5 kotlin SDK")
+        packaging = "pom"
+        name = "Web5 SDK for the JVM"
+        description.set("Web5 SDK for the JVM")
         url.set("https://github.com/TBD54566975/web5-kt")
         inceptionYear.set("2023")
         licenses {
--- a/credentials/build.gradle.kts
+++ b/credentials/build.gradle.kts
@@ -11,31 +11,43 @@ repositories {
   maven("https://repository.jboss.org/nexus/content/repositories/thirdparty-releases/")
 }
 
-val ktor_version = "2.3.4"
-val jackson_version = "2.14.2"
-
 dependencies {
-  api("com.danubetech:verifiable-credentials-java:1.6.0")
+  /**
+   * Maintainers - please do not declare versioning here at the module level;
+   * versioning is centralized for the platform in $projectRoot/gradle/libs.versions.toml
+   *
+   * Deps are declared in alphabetical order.
+   */
 
+  // API
+  /*
+   * API Leak: https://github.com/TBD54566975/web5-kt/issues/228
+   *
+   * Change and move to "implementation" when completed
+   */
+  api(libs.comDanubetechVerifiableCredentials)
+
+  // Project
   implementation(project(":dids"))
   implementation(project(":common"))
   implementation(project(":crypto"))
-  implementation("com.fasterxml.jackson.module:jackson-module-kotlin:$jackson_version")
-  implementation("com.nfeld.jsonpathkt:jsonpathkt:2.0.1")
-  implementation("com.nimbusds:nimbus-jose-jwt:9.34")
-  implementation("decentralized-identity:did-common-java:1.9.0")
-  implementation("com.networknt:json-schema-validator:1.0.87")
 
-  implementation("io.ktor:ktor-client-core:$ktor_version")
-  implementation("io.ktor:ktor-client-okhttp:$ktor_version")
-  implementation("io.ktor:ktor-client-content-negotiation:$ktor_version")
-  implementation("io.ktor:ktor-serialization-jackson:$ktor_version")
-  implementation("io.ktor:ktor-serialization-kotlinx-json:$ktor_version")
-  implementation("io.ktor:ktor-client-logging:$ktor_version")
+  // Implementation
+  implementation(libs.comFasterXmlJacksonModuleKotlin)
+  implementation(libs.comNetworkntJsonSchemaValidator)
+  implementation(libs.comNfeldJsonpathkt)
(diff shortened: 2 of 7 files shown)


ALRubinger:
`SNAPSHOT` deployed to test consumption in `tbdex-kt`: https://s01.oss.sonatype.org/content/repositories/snapshots/xyz/block/web5/0.11.0.issue226-SNAPSHOT/

ALRubinger:
Validated this `SNAPSHOT` in the acceptance testsuite from `developer.tbd.website`: https://github.com/TBD54566975/developer.tbd.website/actions/runs/7868808932/job/21466693595?pr=1226

Ready. Requesting priority review and merge to `main` to prevent rebase errors as forward development continues. After this, will need a release of `web5-kt` and upgrades to https://github.com/TBD54566975/tbdex-kt/pull/160, then a release of `tbdex-kt` there.

mistermoe:
Looks very straightforward @ALRubinger thanks for digging into catalogs and figuring it out! 

qq: Will others be able to consume this catalog outside of this project (e.g. tbdex-kt) as a means to keep both projects shared dependency versions aligned with a single source of truth?

credentials/build.gradle.kts
```diff
@@ -11,31 +11,43 @@ repositories {
...
+   * Deps are declared in alphabetical order.
+   */
+
+  // API
+  /*
+   * API Leak: https://github.com/TBD54566975/web5-kt/issues/228
+   *
+   * Change and move to "implementation" when completed
+   */
+  api(libs.comDanubetechVerifiableCredentials)
```
diehuxx, line 28: Is this is bad because it's a security issue? Or because is it a license-related thing? or both

credentials/build.gradle.kts
```diff
@@ -11,31 +11,43 @@ repositories {
...
+  // Test
+  /**
+   * Test dependencies may declare direct versions; they are not exported
+   * and therefore are within the remit of this module to self-define
+   * if desired.
+   */
+  testImplementation(libs.ioKtorClientMock)
   testImplementation(kotlin("test"))
-  testImplementation("com.willowtreeapps.assertk:assertk:0.27.0")
+  testImplementation(libs.comWillowtreeappsAssertk)
```
diehuxx, line 51: Is the more object-y import better than the string-y import for a specific reason or is it a style thing?

crypto/build.gradle.kts
```diff
@@ -7,19 +7,43 @@ repositories {
...
-val jackson_version = "2.14.2"
-
 dependencies {
-  api("com.nimbusds:nimbus-jose-jwt:9.34")
-  implementation("com.google.crypto.tink:tink:1.10.0")
-  implementation("org.bouncycastle:bcprov-jdk15to18:$bouncy_castle_version")
-  implementation("org.bouncycastle:bcpkix-jdk15to18:$bouncy_castle_version")
+
+  /**
+   * Maintainers - please do not declare versioning here at the module level;
```
diehuxx, line 13: i really appreciate the detailed comments here and in other files

credentials/build.gradle.kts
jiyoontbd: i _think_ this is referenced via `gradle/libs.versions.toml` ? then we can use the same object-y import (with same version) for each subproject (if we use it in more than one subproject)

diehuxx (approved):
jvm packing is not an area of expertise for me, so i'm eager for others to weigh in too. But to me this looks good and really informative. Thanks @ALRubinger for getting this cleaned up!

jiyoontbd (commented):
it would be neato if you could add a little blurb on the main readme under `Development` heading about how we manage dependencies throughout each subproject.

ALRubinger:
> Looks very straightforward @ALRubinger thanks for digging into catalogs and figuring it out!

🤘🏻
 
> qq: Will others be able to consume this catalog outside of this project (e.g. tbdex-kt) as a means to keep both projects shared dependency versions aligned with a single source of truth?

Unfortunately not yet. I can incrementally do follow-on PRs to do:

* Publishing of the version catalog as a separate module and try consuming that in `tbdex-kt`
* Rewriting the entire build in Maven, likely after tbDEX 1.0

I'm having a call with a super expert on this, likely this week. Discovered in this work a whole world of "aligning Gradle dep resolution and how it generates Maven POMs" and the integration those experts suggest would require more structural changes to `web5-kt` and `tbdex-kt` than I wanted to surface to the team right now - so they'd stay free to focus on SDK work and we don't have to worry about chonker refactors.

build.gradle.kts
```diff
@@ -240,14 +241,14 @@ publishing {
...
-      description = "Kotlin SDK for web5 functionality"
+      description = "Web5 SDK for the JVM"
       version = project.property("version").toString()
       from(components["java"])
 
       pom {
-        packaging = "jar"
-        name = project.name
-        description.set("web5 kotlin SDK")
+        packaging = "pom"
```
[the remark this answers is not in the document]
ALRubinger: It maps to the `<packaging> element of the generated `pom.xml` to be of type `pom`, which reflects that this parent module does not have an associated `artifact.jar` as part of its publishing (as it doesn't)

You can see an example of the generated POM [here](https://s01.oss.sonatype.org/content/repositories/snapshots/xyz/block/tbdex/0.11.0.issue159-SNAPSHOT/tbdex-0.11.0.issue159-20240212.074437-3.pom)

credentials/build.gradle.kts
ALRubinger: Exactly, @jiyoontbd! Only reason I didn't enforce the same objectey format in test deps throughout the subprojects is because test deps don't *need* to be aligned among modules; they're not exported to the compile or runtime classpaths of consuming projects.

dids/build.gradle.kts
```diff
@@ -11,33 +11,46 @@ repositories {
...
+  // Test
+  /**
+   * Test dependencies may declare direct versions; they are not exported
+   * and therefore are within the remit of this module to self-define
+   * if desired.
+   */
   testImplementation(kotlin("test"))
-  testImplementation("io.ktor:ktor-client-mock:$ktor_version")
+  testImplementation(libs.ioKtorClientMock)
   testImplementation("org.mockito.kotlin:mockito-kotlin:5.1.0")
```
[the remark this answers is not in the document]
ALRubinger: Up to you! For test deps, I'm leaving it to the project devs to determine if you'd like to uplevel into `libs.versions.toml`. I left them like this because as test deps, they're not exported to the runtime or compile `classpath`s of consumers, and these are therefore with the remit of subprojects to either define for themselves or share.

gradle/libs.versions.toml
```diff
@@ -0,0 +1,83 @@
...
+com_fasterxml_jackson_module = "2.14.2"
+com_github_multiformats = "1.1.0"
+com_google_crypto_tink = "1.10.0"
+com_networknt_json_schema_validator = "1.0.87"
+com_nfeld_jsonpathkt = "2.0.1"
+com_nimbusds = "9.34"
+com_squareup_okhttp3 = "4.12.0"
+com_willowtreeapps_assertk = "0.27.0"
+decentralized_identity_did_common_java = "1.9.0"
+dnsjava = "3.5.2"
```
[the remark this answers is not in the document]
ALRubinger: `dnsjava` was the `groupId`; vs something like a `groupId:artifactId` of `decentralized-identity:did-common-java` which I've mapped with underscores because of the `.toml` referencing format.

credentials/build.gradle.kts
ALRubinger: Great question!

API Leaks are not necessarily bad - they could be the intended behaviour. In these cases I've vetted the [inventory](https://github.com/TBD54566975/web5-kt/issues?q=is%3Aissue+is%3Aopen+%22API+Leak%22) with @mistermoe and he indicated to me he'd like to address them.

An API leak means that another project's (let's call that Project X) API is exposed through yours. This has the following implications:

* The consumer of your project needs Project X on its compilation `classpath`
* Project X may bring in a bunch of other dependencies transitively. 
* The more dependencies you require:
  * Higher risk of something in that supply chain having a security vulnerability
  * Higher risk of something in that supply chain using an incompatible license
  * Increased burden on us as project devs to manage the potential version conflicts in a big dependency tree
* You're letting ProjectX's API be part of your project's API surface area. So if they change their API - yours does too, and all consumers must update. You lose authority over locking your API.

in short, API leaks break the software design principle of _encapsulation_, which isn't *necessarily* bad but is something to be inspected on a case by case basis to see if there's a valid reason. In our cases I've inventoried, I'm hearing from the team - nope, let's fix this up and align the Kotlin SDKs with approaches with taken in other language impls.

ALRubinger:
@jiyoontbd I added some `README` docs as you suggested in fc7df875ec7ca146f17621af1a0581c31c4008cf.

jiyoontbd (approved):

ALRubinger:
@jiyoontbd @mistermoe is this clear for me to rebase/squash/merge? I see 2 approving reviews and also this:

image

So I can do it but would be bypassing branch protections to do so.

mistermoe (approved):

mistermoe:
@ALRubinger fire away

PR #237, opened 2024-02-13
Update did:dht resolution on invalid uri

nitro-neal:
# Overview
Update did:dht resolution

# Description
Updates did:dht resolution error to be in line with our js implementation and test vectors

When an invalid did uri is given to the dht resolver it should come back with `invalidDid` error


decentralgabe (approved):

dids/src/main/kotlin/web5/sdk/dids/methods/dht/DidDht.kt
```diff
@@ -241,12 +241,14 @@ public sealed class DidDhtApi(configuration: DidDhtConfiguration) : DidMethod<Di
   private fun resolveInternal(did: String): DidResolutionResult {
     try {
       validate(did)
+    } catch (_: ParserException) {
```
jiyoontbd, line 244: could we write a test for this?

jiyoontbd:
@nitro-neal 👋 would you like to add a test for the parserexception (i think just adding one more test case to `resolve.json` in web5-spec submodule?) and then we can go ahead and merge this smol pr

nitro-neal:
Added unit test 👍

diehuxx (approved):

PR #241, opened 2024-02-14
Suite of Security Vulnerability Fixes

### LINKED ISSUE #240: Apply forced version resolution to all projects
Forced version resolution is a step we take to manually upgrade transitive dependencies with CVEs. Right now we're declaring them in the parent build file; we need these to apply to all projects so that the testsuites for the subprojects run in the right `classpath`.

Sets the stage for our applying further security vulnerability fixes.

ALRubinger:
A series of commits to address security vulns in the `web5-kt` projects. Grouped as one PR so they may be tested together to clear FOSSA security scanning to  ✅. 

On `main`, 5 security issues on FOSSA:
image

In this PR, 0 security issues on FOSSA:
image

Recommended merge approach when approved: Rebase, do not squash and merge. This will preserve optionality for us to individually `git blame` and do reverts if necessary.

This PR resolves #240, #242, #243, #244, and adds documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,12 @@ versions in `gradle/libs.versions.toml` and import references defined
 there in the subproject `build.gradle.kts` files. More docs on this 
 approach using Gradle Version Catalogs is at the top of `gradle/libs.versions.toml`.
 
+We have a secondary mechanism to force dependency upgrades of transitive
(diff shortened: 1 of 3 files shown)


jiyoontbd (approved):
🙏

michaelneale (approved):
this is very good.

PR #251, opened 2024-02-25
FIX: `did:dht` resolution

mistermoe:
# Overview
fixed `did:dht` resolution.

# Description
There was a recent change to `did:dht` TXT record names that requires the method specific id to be included as the last part of each TXT record

image

`main` is currently looking for several records using strict equality. example [here](https://github.com/TBD54566975/web5-kt/blob/main/dids/src/main/kotlin/web5/sdk/dids/methods/dht/DidDht.kt#L572). 

This PR changes strict equality checks to prefix checks. We would have had to make this change for the root record anyway because of the possibility for `_did.TLD`

> [!WARNING]
> I'll fix create and update in a subsequent PR. want to get this one out to unblock others

### DIFF
--- a/dids/src/main/kotlin/web5/sdk/dids/methods/dht/DidDht.kt
+++ b/dids/src/main/kotlin/web5/sdk/dids/methods/dht/DidDht.kt
@@ -559,7 +559,7 @@ public sealed class DidDhtApi(configuration: DidDhtConfiguration) : DidMethod<Di
                 .build()
             }
             // handle type indexing
-            name == "_typ._did." -> {
+            name.startsWith("_typ._did.") -> {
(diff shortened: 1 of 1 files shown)


mistermoe:
@jiyoontbd lmk what you'd like to do about codecov percentages! looks like overall test coverage went up but i somehow didn't hit the required % diff for a patch? went and looked at the lines that aren't covered aaand they have nothing to do with the changes made in this PR

michaelneale:
yeah this looks like the codecov thing is not relevant to this patch.

michaelneale (approved):

phoebe-lew (approved):

jiyoontbd (approved):
cool! we just read this code earlier today heh

Issue #217, opened 2024-02-05
Add Maven Build


ALRubinger:
Implement a Maven build; from some team notes we'd like to move to this build system for clarity and inter-project consistency.

This process will also lay bare the necessary auditing of dependencies to align versions.

WIP in PR #261 and `upstream` branch `issue-217/maven-build`

ALRubinger:
WIP on my branch: https://github.com/ALRubinger/web5-kt/tree/issue-217/maven-build

ALRubinger:
WIP now in an upstream branch for sharing with others: https://github.com/TBD54566975/web5-kt/tree/issue-217/maven-build

ALRubinger:
Closing this, going to address version alignment in the existing Gradle builds in #226

ALRubinger:
Re-opening; needed for version alignment between Web5 and tbDEX

ALRubinger:
Done in 89fc4539846c5d39b842460a7256ec005488bb3b

PR #256, opened 2024-03-04
 Automate GH Release Publication

leordev:
Similar to https://github.com/TBD54566975/tbdex-kt/pull/188

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,28 +111,39 @@ To build and run test just run:
 gradle build --console=rich
 ```
 
-## Releasing
+## Release Guidelines
 
-In order to release to Central Repository, simply cut a new release tag in the Github UI. There is a configured [Github
-Action](./.github/workflows/publish.yml) that will automatically publish the release to Central Repository. You can cut
-the release by going to the [create a new releases page](https://github.com/TBD54566975/web5-kt/releases/new). When
-creating a new tag, the name should be in the format `vX.Y.Z`. Please note that once a release is made, it is immutable
-and cannot be deleted.
+### Pre-releases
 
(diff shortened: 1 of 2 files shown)


KendallWeihe (approved):

README.md
```diff
@@ -111,28 +111,39 @@ To build and run test just run:
...
+
+2. In the version field, insert the current version, a short meaningful identifier and the `-SNAPSHOT` prefix, ie:
+
+   - 0.11.0.pr123-SNAPSHOT
+   - 0.11.0.shortsha-SNAPSHOT
+   - 0.11.0.fixsomething-SNAPSHOT
+
+3. Run workflow!
+
+**DON'T FORGET THE `-SNAPSHOT` SUFFIX**, otherwise it will generate publish a new official release to maven registry.
```
KendallWeihe, line 132: Hmmm I wonder if maybe we should implement something to prevent accidents here. Like, maybe we have a dedicated workflow for pre-release, even though everything would be the same as a production release, at least the developer would be more intentional from the page load? Idk I still don't love that idea... what could we do? 🤔

KendallWeihe: Maybe somehow `SNAPSHOT` would be the default and it would require developer intervention to override it?

ALRubinger: @KendallWeihe - This is a great point; @leordev @finn-tbd and I are going to handle this as part of #217 . There will then be 2 workflows: one for SNAPSHOTs and one for releases. Each workflow will perform precondition checks on the user input for version to make sure the `-SNAPSHOT` suffix is either present or not depending upon the workflow.

PR #262, opened 2024-03-07
Implementing BearerDid / PortableDid, exposing our own JWS, JWT, JWK types

jiyoontbd:
# Motivation
* Align public API surface of DID creation and credential signing to be consistent with other web5 repos

# Overview
* `BearerDid` and `PortableDid` types used in place of `Did(uri, keyManager)` abstract class. Now all existing DID implementations' `DidXyz.create()` to return `BearerDid`.
* Our own `Jws`, `Jwt`, `Jwk` types are used in place of nimbusds types. We are still creating private key internally using nimbusds (see `Ed25519.kt` for example), but all public surface APIs no longer expose the third party dependency.

# How Has This Been Tested?
* Ensured existing tests compiled and passed 
* Wrote new tests for new types and methods

# Checklist

- [x] I have read the CONTRIBUTING document.
- [x] My code is consistent with the rest of the project 
- [x] I have tagged the relevant reviewers and/or interested parties
- [x] I have updated the READMEs and other documentation of affected packages


dids/src/main/kotlin/web5/sdk/dids/did/BearerDID.kt
```diff
@@ -0,0 +1,89 @@
...
+      uri = this.did.uri,
+      document = this.document,
+      privateKeys = privateKeys,
+      metadata = mapOf()
+    )
+  }
+
+  public companion object {
+
+    public fun import(
```
jiyoontbd, line 59: @mistermoe would like to know what you think of this - currently i've implemented import here, and then am calling it `BearerDID.import(..)` in didjwk, dht, etc.

the implementation is _slightly_ different across all web5-sdks but i've tried to adhere to the JS one the most, while referencing swift and go sdks.

dids/src/main/kotlin/web5/sdk/dids/methods/dht/DidDht.kt
```diff
@@ -136,7 +135,7 @@ public sealed class DidDhtApi(configuration: DidDhtConfiguration) : DidMethod<Di
     val publicKey = keyManager.getPublicKey(keyAlias)
 
     // build DID Document
-    val id = DidDht.getDidIdentifier(publicKey)
+    val didUrl = DidDht.getDidIdentifier(publicKey)
```
jiyoontbd, line 138: todo probably should find out what this method returns and call it something other than `didUrl` since it's probably not the full did

jiyoontbd: changed to `didUri`

dids/src/main/kotlin/web5/sdk/dids/extensions/DidExtensions.kt
```diff
@@ -20,6 +20,6 @@ internal val supportedMethods = mapOf(
  * signing and managing the passed in [didUri] exists within the provided [keyManager]. This function is meant
  * to be used when the method of the DID is unknown.
  */
-public fun Did.Companion.load(didUri: String, keyManager: KeyManager): Did {
-  return supportedMethods.getValue(DidUri.parse(didUri).method).load(didUri, keyManager)
+public fun ChangemeDid.Companion.load(didUri: String, keyManager: KeyManager): ChangemeDid {
```
jiyoontbd, line 23: @mistermoe tagging to look into how to remove this method async

crypto/src/main/kotlin/web5/sdk/crypto/AwsKeyManager.kt
```diff
@@ -165,11 +168,11 @@ public class AwsKeyManager @JvmOverloads constructor(
...
    * Return the alias of [publicKey], as was originally returned by [generatePrivateKey].
    *
-   * @param publicKey A public key in JWK (JSON Web Key) format
+   * @param publicKey A public key in Jwk (JSON Web Key) format
    * @return The alias belonging to [publicKey]
    */
-  override fun getDeterministicAlias(publicKey: JWK): String {
-    val jwkThumbprint = publicKey.computeThumbprint()
+  override fun getDeterministicAlias(publicKey: Jwk): String {
+    val jwkThumbprint = publicKey.kid ?: publicKey.computeThumbprint()
```
mistermoe, line 175: i think this should just be `publicKey.computeThumbprint()` bc `kid` may not be the thumbprint which makes the alias non-deterministic

mistermoe, line 175: i think this should just be `publicKey.computeThumbprint()` bc `kid` may not be the thumbprint which makes the alias non-deterministic

crypto/src/main/kotlin/web5/sdk/crypto/jwk/Jwk.kt
```diff
@@ -0,0 +1,180 @@
...
+ * @property y Y coordinate for EC keys.
+ *
+ */
+public class Jwk(
+  public val kty: String,
+  public val crv: String,
+  public val use: String?,
+  public val alg: String?,
+  public var kid: String?,
+  public val d: String? = null,
```
mistermoe, line 39: is there a reason we have to explicitly set this to `null`?

dids/src/main/kotlin/web5/sdk/dids/did/BearerDid.kt
```diff
@@ -0,0 +1,143 @@
...
+   * a DID in format that can be used independently of the specific DID method implementation. It
+   * extracts both public and private keys from the DID's key manager and organizes them into a
+   * `PortableDid` structure.
+   *
+   * @returns A `PortableDid` containing the URI, DID document, metadata, and optionally private
+   *          keys associated with the `BearerDid`.
+   */
+  public fun export(): PortableDid {
+
+    val keyExporter = keyManager as? KeyExporter
```
mistermoe, line 77: we should do an `is KeyExporter` check and only run the exports if the keymanager is an exporter.  otherwise ignore and proceed

dids/src/main/kotlin/web5/sdk/dids/did/BearerDid.kt
```diff
@@ -0,0 +1,143 @@
...
+      check(allVerificationMethodsHavePublicKey) {
+        "Each VerificationMethod must contain a public key in Jwk format."
+      }
+
+      val did = Did.parse(portableDid.uri)
+
+      for (key in portableDid.privateKeys) {
+        val keyImporter = keyManager as? KeyImporter
+        keyImporter!!.importKey(key)
+      }
```
mistermoe, line 136: `if (keyManager is KeyImporter)`

common/src/main/kotlin/web5/sdk/common/Convert.kt
```diff
@@ -46,7 +46,7 @@ public val B64URL_DECODER: Base64.Decoder = Base64.getUrlDecoder()
  * ```
  * // Example 1: Convert a ByteArray to a Base64Url encoded string without padding
  * val byteArray = byteArrayOf(1, 2, 3)
- * val base64Url = Convert(byteArray).toBase64Url(padding = false)
+ * val base64Url = Convert(byteArray).toBase64Url()
```
nitro-neal, line 49: should we put the padding = false back in the documentation ?

credentials/src/main/kotlin/web5/sdk/credentials/PresentationExchange.kt
```diff
@@ -161,12 +164,16 @@ public object PresentationExchange {
...
-      val vc = JWTParser.parse(vcJwt) as SignedJWT
+    val vcJwtListWithNodes = vcJwtList.zip(
+      vcJwtList.map { vcJwt ->
+        val vc = Jwt.decode(vcJwt)
 
-      JsonPath.parse(vc.payload.toString())
-        ?: throw JsonPathParseException()
-    })
+        val jwtModule = SimpleModule().addSerializer(JwtClaimsSet::class.java, JwtClaimsSetSerializer())
+        Json.jsonMapper.registerModule(jwtModule)
```
diehuxx, line 172: Looks like we create and register the same `jwtModule` on every iteration of the `.map`. Maybe move this above the `.map` so we only do it once?

common/src/main/kotlin/web5/sdk/common/Convert.kt
jiyoontbd: i removed it coz padding is now false by default and added language around that default value, but can add back in just to show the usage?

crypto/src/main/kotlin/web5/sdk/crypto/Dsa.kt
```diff
@@ -12,6 +12,21 @@ public enum class JwaCurve {
   Ed25519;
 
   public companion object {
+
+    /**
+     * Convert JwaCurve nimbusds JWK curve.
+     * Used to temporarily bridge the gap between moving from nimbusds JWK methods
+     * to rolling our own JWK methods
+     * @param curve
+     * @return nimbus JWK Curve
```
diehuxx, line 21: clarification: I thought we're trying not to return nimbus objects. Is this an exceptional case or i have I misunderstood the goal?

credentials/src/main/kotlin/web5/sdk/credentials/VerifiableCredential.kt
```diff
@@ -64,15 +63,15 @@ public class VerifiableCredential internal constructor(public val vcDataModel: V
...
-  public fun sign(did: Did, assertionMethodId: String? = null): String {
-    val payload = JWTClaimsSet.Builder()
+  public fun sign(did: BearerDid, assertionMethodId: String? = null): String {
+    val payload = JwtClaimsSet.Builder()
       .issuer(vcDataModel.issuer.toString())
-      .issueTime(vcDataModel.issuanceDate)
+      .issueTime(vcDataModel.issuanceDate.time)
       .subject(vcDataModel.credentialSubject.id.toString())
-      .claim("vc", vcDataModel.toMap())
+      .misc("vc", vcDataModel.toMap())
```
nitro-neal, line 71: is `misc` the more official name than `claim`?

credentials/src/test/kotlin/web5/sdk/credentials/PresentationExchangeTest.kt
```diff
@@ -69,8 +70,8 @@ class PresentationExchangeTest {
       )
       val vc = VerifiableCredential.create(
         type = "StreetCred",
-        issuer = issuerDid.uri,
-        subject = holderDid.uri,
+        issuer = issuerDid.did.uri,
```
nitro-neal, line 73: just fyi in ts it's `issuerDid.uri`

nitro-neal:
```
export class BearerDid {
  /** {@inheritDoc Did#uri} */
  uri: string;
  ...
  }

File: dids/src/main/kotlin/web5/sdk/dids/DidMethod.kt [hunk 13]
```diff
@@ -64,121 +32,3 @@ public interface CreateDidOptions
  * ```
  */
 public interface CreationMetadata
```
diehuxx, line 34: Looks like `CreationMetadata` is only used with did:ion, which we're dropping support for anyway, right? Is it worth dropping ion in this PR?

dids/src/test/kotlin/web5/sdk/dids/methods/key/DidKeyTest.kt
```diff
@@ -24,52 +21,25 @@ class DidKeyTest {
     @Test
     fun `it works`() {
```
diehuxx, line 22: nit: slightly more descriptive test name please
```suggestion
    fun `creates and resolves did:key`() {
```

crypto/src/main/kotlin/web5/sdk/crypto/jwk/Jwk.kt
jiyoontbd: nope, not sure why i put it there. we use the builder pattern anyway and don't use this constructor

dids/src/main/kotlin/web5/sdk/dids/did/BearerDid.kt
jiyoontbd: by `ignore and proceed`, do you mean we should throw an exception, or something else? like have this check at the top

```kotlin
check(keyManager is KeyExporter) {
      "KeyManager must implement KeyExporter to export keys"
    }
```

dids/src/main/kotlin/web5/sdk/dids/did/BearerDid.kt
jiyoontbd: i put a check at the top `check(keyManager is KeyImporter) { "error" }`

credentials/src/main/kotlin/web5/sdk/credentials/PresentationExchange.kt
jiyoontbd: great call, thanks!

crypto/src/main/kotlin/web5/sdk/crypto/Dsa.kt
jiyoontbd: this is an exceptional case, thanks for bringing it up!

currently, we still have to use nimbusds internally to create the private key material. there's a follow-up issue to completely eject from nimbusds and write the private key ourselves by using google tink library: https://github.com/TBD54566975/web5-kt/issues/273

this method is still used in `AwsKeyManager#getPublicKey`, where we take `java.security.PublicKey` and convert it to an ECPublicKey, which is one way we create privatekey using nimbusds library. 

we "ejected from nimbusds" in this PR in the sense that we no longer expose nimbusds' JWK type as part of the library API surface.

i expect that we can safely get rid of this method once we work on that follow-up issue.

credentials/src/main/kotlin/web5/sdk/credentials/VerifiableCredential.kt
jiyoontbd: `misc` is the top level field where we shove everything that's not a reserved keyword in JWT. i've followed the implementation in web5-dart

https://github.com/TBD54566975/web5-dart/blob/main/packages/web5/lib/src/jwt/jwt_claims.dart#L59

@mistermoe is there a spec i can reference for neal / do we need to write one?

credentials/src/test/kotlin/web5/sdk/credentials/PresentationExchangeTest.kt
jiyoontbd: ooooooh tank u! i missed that bit 🙏  added `uri` to BearerDid and changed all the places where i'm calling `bearerdid.did.uri` to `bearerdid.uri`

dids/src/main/kotlin/web5/sdk/dids/DidMethod.kt
jiyoontbd: aw yea i'm deleting this, ty!

dids/src/test/kotlin/web5/sdk/dids/methods/key/DidKeyTest.kt
jiyoontbd: done!

credentials/src/main/kotlin/web5/sdk/credentials/VerifiablePresentation.kt
```diff
@@ -58,14 +57,14 @@ public class VerifiablePresentation internal constructor(public val vpDataModel:
...
   @JvmOverloads
-  public fun sign(did: Did, assertionMethodId: String? = null): String {
-    val payload = JWTClaimsSet.Builder()
-      .issuer(did.uri)
-      .issueTime(Date())
-      .claim("vp", vpDataModel.toMap())
+  public fun sign(bearerDid: BearerDid, assertionMethodId: String? = null): String {
+    val payload = JwtClaimsSet.Builder()
+      .issuer(bearerDid.uri)
+      .issueTime(Date().time)
```
nitro-neal, line 63: if this is iat may need to change to seconds, currently this is milliseconds I think

crypto/src/main/kotlin/web5/sdk/crypto/Ed25519.kt
```diff
@@ -135,16 +147,16 @@ public object Ed25519 : KeyGenerator, Signer {
...
    * a descriptive error message.
    *
-   * @param key The [JWK] to validate.
+   * @param key The [Jwk] to validate.
    * @throws IllegalArgumentException if the key is not a public key
    */
-  public fun validatePublicKey(key: JWK) {
-    require(!key.isPrivate) { "key must be public" }
+  public fun validatePublicKey(key: Jwk) {
+    require(key.d == null) { "key must be public" }
```
nitro-neal, line 154: I think in a kt util package we have isPrivateKey() function

dids/src/main/kotlin/web5/sdk/dids/did/BearerDid.kt
```diff
@@ -0,0 +1,153 @@
...
+      val keyAliasResult = runCatching { vm.publicKeyJwk?.computeThumbprint() }
+      if (keyAliasResult.isSuccess) {
+        val keyAlias = keyAliasResult.getOrNull()
+        keyExporter.exportKey(keyAlias!!.toString()).let { key ->
+          privateKeys.add(key)
+        }
+      }
+    }
+
+    return PortableDid(
```
nitro-neal, line 95: is there a unit test that makes sure the output of this portable did matches the js and/or golang one?

jose/src/main/kotlin/web5/sdk/jose/jws/Jws.kt
```diff
@@ -0,0 +1,275 @@
...
+ */
+public object Jws {
+
+  /**
+   * Decode a JWS into its parts.
+   *
+   * @param jws The JWS to decode
+   * @return DecodedJws
+   */
+  @Suppress("SwallowedException")
```
nitro-neal, line 26: this is expected?

jose/src/main/kotlin/web5/sdk/jose/jwt/Jwt.kt
```diff
@@ -0,0 +1,240 @@
...
+ * @property jti provides a unique identifier for the JWT.
+ * @property misc additional claims (i.e. VerifiableCredential, VerifiablePresentation)
+ */
+@JsonSerialize(using = JwtClaimsSetSerializer::class)
+@JsonDeserialize(using = JwtClaimsSetDeserializer::class)
+public class JwtClaimsSet(
+  public val iss: String? = null,
+  public val sub: String? = null,
+  public val aud: String? = null,
+  public val exp: Long? = null,
```
nitro-neal, line 142: I think long is the right impl here but I think i see other libs with int

nitro-neal (approved):
Wow great job ! Lots of hard work here.

mistermoe (approved):

diehuxx (approved):
great work!

credentials/src/main/kotlin/web5/sdk/credentials/VerifiableCredential.kt
decentralgabe: the standard to use is public / private claim names, depending on whether the claims are registered in an IANA registry

https://datatracker.ietf.org/doc/html/rfc7519#section-4.2

decentralgabe: I think `claims` is more apt than `misc`

credentials/src/test/kotlin/web5/sdk/credentials/VerifiablePresentationTest.kt
```diff
@@ -161,19 +161,20 @@ class VerifiablePresentationTest {
     val keyManager = InMemoryKeyManager()
     val holderDid = DidKey.create(keyManager)
 
-    val header = JWSHeader.Builder(JWSAlgorithm.ES256K)
-      .keyID(holderDid.uri)
+    val header = JwsHeader.Builder()
+      .type("JWT")
+      .algorithm(Jwa.ES256K.name)
+      .keyId(holderDid.uri)
```
decentralgabe, line 167: noting that the key ID needs to be a did plus a key reference (e.g. `did:example:abcd#key-1`)

crypto/src/main/kotlin/web5/sdk/crypto/Crypto.kt
```diff
@@ -74,11 +74,11 @@ public object Crypto {
...
-   * @return The computed public key as a JWK object.
+   * @return The computed public key as a Jwk object.
    */
-  public fun computePublicKey(privateKey: JWK): JWK {
-    val rawCurve = privateKey.toJSONObject()["crv"]
-    val curve = rawCurve?.let { JwaCurve.parse(it.toString()) }
+  public fun computePublicKey(privateKey: Jwk): Jwk {
+    val rawCurve = privateKey.crv
+    val curve = JwaCurve.parse(rawCurve)
     val generator = getKeyGenerator(AlgorithmId.from(curve))
```
decentralgabe, line 82: calling out that this assumes we are only using keys that use ECC....which is fine for now but may break at some point in the future (if we need to use something like RSA or post quantum)

crypto/src/main/kotlin/web5/sdk/crypto/Ed25519.kt
```diff
@@ -43,67 +42,80 @@ public object Ed25519 : KeyGenerator, Signer {
...
+        privateKey.x?.let { x(it) }
+      }
+      .build()
 
-    return privateKey.toOctetKeyPair().toPublicJWK()
+    return jwk
   }
 
-  override fun privateKeyToBytes(privateKey: JWK): ByteArray {
+  override fun privateKeyToBytes(privateKey: Jwk): ByteArray {
```
decentralgabe, line 81: it would be good to have methods that are able to represent compressed keys

dids/src/main/kotlin/web5/sdk/dids/methods/key/DidKey.kt
```diff
@@ -144,15 +136,15 @@ public class DidKey(uri: String, keyManager: KeyManager) : Did(uri, keyManager)
 
       val publicKeyJwk = keyGenerator.bytesToPublicKey(publicKeyBytes)
 
-      val verificationMethodId = "${parsedDidUri.uri}#$id"
+      val verificationMethodId = "${parsedDid.uri}#$id"
       val verificationMethod = VerificationMethod.Builder()
         .id(verificationMethodId)
         .publicKeyJwk(publicKeyJwk)
         .controller(did)
         .type("JsonWebKey2020")
```
decentralgabe, line 144: if we're using `JsonWebKey` with `did:jwk` we should use it here too

dids/src/test/kotlin/web5/sdk/dids/methods/jwk/DidJwkTest.kt
```diff
@@ -12,19 +10,22 @@ import web5.sdk.common.Convert
...
 import kotlin.test.assertEquals
 import kotlin.test.assertNotNull
-import kotlin.test.assertTrue
 
 class DidJwkTest {
   @Nested
   inner class CreateTest {
     @Test
-    fun `creates an ES256K key when no options are passed`() {
+    fun `creates an EdDSA key when no options are passed`() {
```
decentralgabe, line 28:
```suggestion
    fun `creates an Ed25519 key when no options are passed`() {
```

EdDSA is an algorithm not a key 😄

jiyoontbd:
a reminder to myself to first merge https://github.com/TBD54566975/web5-spec/pull/130 and then change the submodule commit to the merge commit

credentials/src/main/kotlin/web5/sdk/credentials/VerifiablePresentation.kt
jiyoontbd: fixed, and updated the ktdoc

jose/src/main/kotlin/web5/sdk/jose/jwt/Jwt.kt
jiyoontbd: i agree it should stay as a long - can you tell me which lib has it as an int so we can write an issue for it? i checked js (number), go(int64) and swift (ISO8601Date)

credentials/src/test/kotlin/web5/sdk/credentials/VerifiablePresentationTest.kt
jiyoontbd: oh yeah! fixed, thank you

dids/src/main/kotlin/web5/sdk/dids/methods/key/DidKey.kt
jiyoontbd: oh yea, this should also be updated! changed to `JsonWebKey`

dids/src/main/kotlin/web5/sdk/dids/did/BearerDid.kt
jiyoontbd: hmm... that sounds like a test vector job. i shall write a test vector!

jiyoontbd: wrote test vector and a test against it: https://github.com/TBD54566975/web5-kt/pull/262/commits/b47299053369c359cb49aad2b2ac6612f36d3419

crypto/src/main/kotlin/web5/sdk/crypto/Crypto.kt
jiyoontbd: 🤯

jiyoontbd: added a comment about this in ktdoc for `getKeyGenerator()` method here: https://github.com/TBD54566975/web5-kt/pull/262/commits/c5bba53b252324e975975f7c6f72edd4647f827c

crypto/src/main/kotlin/web5/sdk/crypto/AwsKeyManager.kt
```diff
@@ -118,24 +117,25 @@ public class AwsKeyManager @JvmOverloads constructor(
...
     val getPublicKeyRequest = GetPublicKeyRequest().withKeyId(keyAlias)
     val publicKeyResponse = kmsClient.getPublicKey(getPublicKeyRequest)
     val publicKey = convertToJavaPublicKey(publicKeyResponse.publicKey)
 
     val algorithmDetails = getAlgorithmDetails(publicKeyResponse.keySpec.enum())
     val jwkBuilder = when (publicKey) {
-      is ECPublicKey -> ECKey.Builder(JwaCurve.toJwkCurve(algorithmDetails.curve), publicKey)
+      is ECPublicKey -> {
+        val key = ECKey.Builder(JwaCurve.toNimbusCurve(algorithmDetails.curve), publicKey).build()
+        Jwk.Builder("EC", key.curve.name)
```
tomdaffurn, line 132: Does this defs populate algorithm, keyID, keyUse?

tomdaffurn (approved):
Approved with comment.

Huge amount of work, well done Jiyoon. Good riddance nimbus

crypto/src/main/kotlin/web5/sdk/crypto/AwsKeyManager.kt
jiyoontbd: `alg` and `kid` is now not populated by default - relevant comment [here](https://github.com/TBD54566975/web5-spec/pull/130#discussion_r1541332623)

`kid` can be computed when needed via `Jwk#computeThumbprint()`

`keyUse` hmm... i'm pretty sure we're also not populating these by default, and it is optional per JWK spec
https://datatracker.ietf.org/doc/html/rfc7517#section-4.2

> Values defined by this specification are:
>
>   o  "sig" (signature)
>   o  "enc" (encryption)
>
>   Other values MAY be used.  The "use" value is a case-sensitive
>   string.  Use of the "use" member is OPTIONAL, unless the application
>   requires its presence.

jiyoontbd: the only fields that are required to build a Jwk are kty and crv

credentials/src/main/kotlin/web5/sdk/credentials/VerifiableCredential.kt
jiyoontbd: leaving it as `misc` for now, but this needs a discussion and a decision - https://github.com/TBD54566975/web5-spec/issues/143

mistermoe: agreed @decentralgabe . `misc` is just something i came up with to avoid having to use the terms `private` and `public` in code because those are reserved. i think @nitro-neal 's `claim()` suggestion is a great idea. can address in a subsequent PR

crypto/src/main/kotlin/web5/sdk/crypto/AwsKeyManager.kt
```diff
@@ -118,24 +117,25 @@ public class AwsKeyManager @JvmOverloads constructor(
...
    * @throws [AWSKMSException] for any error originating from the [AWSKMS] client
    */
-  override fun getPublicKey(keyAlias: String): JWK {
+  override fun getPublicKey(keyAlias: String): Jwk {
     val getPublicKeyRequest = GetPublicKeyRequest().withKeyId(keyAlias)
     val publicKeyResponse = kmsClient.getPublicKey(getPublicKeyRequest)
     val publicKey = convertToJavaPublicKey(publicKeyResponse.publicKey)
 
     val algorithmDetails = getAlgorithmDetails(publicKeyResponse.keySpec.enum())
     val jwkBuilder = when (publicKey) {
```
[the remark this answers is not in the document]
mistermoe: good question not sure. it does handle converting what aws returns into a JWK correctly

crypto/src/main/kotlin/web5/sdk/crypto/Ed25519.kt
mistermoe: for `Ed25519` @decentralgabe ? or are you talking about `Secp256k1` or more generally `EC`?

jiyoontbd: would like gabe to weigh in on this / edit this issue to clarify what's missing here

https://github.com/TBD54566975/web5-kt/issues/280

PR #275, opened 2024-03-20
Fix OSSF scorecard action

finn-tbd:
This fixes the scorecard CI action that will likely fail when next run if not updated


ALRubinger (approved):
Applying across the full suite

jiyoontbd (approved):

PR #291, opened 2024-04-01
Make VC Verify more robust

nitro-neal:
# Overview
Added vcJwt checks to verify

# Description
     * If any of these steps fail, the function will throw a [IllegalArgumentException] with a message indicating the nature of the failure:
     * - exp MUST represent the expirationDate property, encoded as a UNIX timestamp (NumericDate).
     * - iss MUST represent the issuer property of a verifiable credential or the holder property of a verifiable presentation.
     * - nbf MUST represent issuanceDate, encoded as a UNIX timestamp (NumericDate).
     * - jti MUST represent the id property of the verifiable credential or verifiable presentation.
     * - sub MUST represent the id property contained in the credentialSubject.

# How Has This Been Tested?
Added new vector - https://github.com/TBD54566975/web5-spec/pull/146

### DIFF
--- a/credentials/src/test/kotlin/web5/sdk/credentials/PresentationExchangeTest.kt
+++ b/credentials/src/test/kotlin/web5/sdk/credentials/PresentationExchangeTest.kt
@@ -664,13 +664,13 @@ class Web5TestVectorsPresentationExchange {
     val testVectors =
       mapper.readValue(File("../web5-spec/test-vectors/presentation_exchange/validate_definition.json"), typeRef)
 
-    testVectors.vectors.filterNot { it.errors ?: false }.forEach { vector ->
+    testVectors.vectors.filter { it.errors == false }.forEach { vector ->
       assertDoesNotThrow {
         PresentationExchange.validateDefinition(vector.input.presentationDefinition)
       }
     }
 
-    testVectors.vectors.filter { it.errors ?: false }.forEach { vector ->
+    testVectors.vectors.filter { it.errors == true }.forEach { vector ->
       assertFails {
           PresentationExchange.validateDefinition(vector.input.presentationDefinition)
       }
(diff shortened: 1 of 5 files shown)


jiyoontbd:
@nitro-neal can you link the spec text where we talk about the date fields like exp iat nbf being numeric values? @kirahsapong and i were wondering about whether it should be a date or a long field while working in web5-swift.

credentials/src/test/kotlin/web5/sdk/credentials/VerifiableCredentialTest.kt
```diff
@@ -331,9 +350,23 @@ class Web5TestVectorsCredentials {
       }
     }
 
-    testVectors.vectors.filter { it.errors ?: false }.forEach { vector ->
-      assertFails {
-        VerifiableCredential.verify(vector.input.vcJwt)
+    testVectors.vectors.filter { it.errors == true }.forEach { vector ->
+      if(vector.errorMessage == null || vector.errorMessage!!["web5-kt"] == null) {
```
jiyoontbd, line 354: i'd rather not have error message be the deciding factor in whether our test is executed. is there another way?

nitro-neal: will update

nitro-neal:
here is vcdata model - https://www.w3.org/TR/vc-data-model/#jwt-encoding

credentials/src/test/kotlin/web5/sdk/credentials/PresentationExchangeTest.kt
```diff
@@ -664,13 +664,13 @@ class Web5TestVectorsPresentationExchange {
     val testVectors =
       mapper.readValue(File("../web5-spec/test-vectors/presentation_exchange/validate_definition.json"), typeRef)
 
-    testVectors.vectors.filterNot { it.errors ?: false }.forEach { vector ->
+    testVectors.vectors.filter { it.errors == false }.forEach { vector ->
```
nitro-neal, line 667: I think intelij is doing this automatically

jiyoontbd (approved):

PR #294, opened 2024-04-03
Update create vector

nitro-neal:
# Overview
Updated the way `VerifiableCredential.create()` works

# Description
The input are the fields for the actual input fields of a vc with the VerifiableCredential.create() method.
The output is the comparison of the VcDataModel object to what is expected.

Example test vector:
```
      "description":"valid simple credential",
      "input":{
        "credential":{
          "type":"SimpleCred",
          "issuer":"did:example:issuer",
          "subject":"did:example:subject",
          "credentialSubject":{
            "legit":true
          }
        }
      },
      "output":{
        "@context":[
          "https://www.w3.org/2018/credentials/v1"
        ],
        "type":[
          "VerifiableCredential",
          "SimpleCred"
        ],
        "id":"urn:uuid:ea74b564-fbe0-4b80-bd60-cd465d433e12",
        "issuer":"did:example:issuer",
        "credentialSubject":{
          "id":"did:example:subject",
          "legit":true
        }
      }
    },
    ```

### DIFF
--- a/credentials/src/test/kotlin/web5/sdk/credentials/VerifiableCredentialTest.kt
+++ b/credentials/src/test/kotlin/web5/sdk/credentials/VerifiableCredentialTest.kt
@@ -398,4 +398,22 @@ class Web5TestVectorsCredentials {
       }
     }
   }
+
+  @Test
+  fun verifyVcJwt() {
+    val typeRef = object : TypeReference<TestVectors<String, Unit>>() {}
+    val testVectors = mapper.readValue(File("../web5-spec/test-vectors/vc_jwt/verify.json"), typeRef)
+
+    testVectors.vectors.filter { it.errors == false }.forEach { vector ->
+      assertDoesNotThrow {
+        VerifiableCredential.verify(vector.input)
+      }
+    }
+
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Review by decentralgabe — APPROVED** [review]

File: credentials/src/test/kotlin/web5/sdk/credentials/VerifiableCredentialTest.kt [hunk 1]
```diff
@@ -316,23 +320,64 @@ class Web5TestVectorsCredentials {
...
+        credentialStatus = CredentialStatus.fromMap(it)
+      }
 
-    testVectors.vectors.filter { it.errors == true }.forEach { vector ->
-      assertFails(vector.description) {
-        VerifiableCredential.fromJson(mapper.writeValueAsString(vector.input.credential))
+      val evidence = vector.input.credential?.get("evidence") as? List<Any>
+
+      val vc = VerifiableCredential.create(
+        type = vector.input.credential?.get("type") as String,
```
jiyoontbd, line 345: there's a lot of `?get("...")` here - have you considered having concrete types for these fields inside vector or credential input?

jiyoontbd (commented):
let's use types! types are cool and fun and neato

Issue #297, opened 2024-04-08
Check out for release using `tbd-releases` account


ALRubinger:
This should allow this account only to bypass the `main` branch protections to push release commits to `main`. This is appropriate only in this case because:

* No code changes are made by the `tbd-releases` account; this is for version management in the `pom.xml` only while releasing and tagging
* Access to the `tbd-releases` account is not shared amongst TBD ENG; constrained to a very small group
* This is only used in the automated release workflow

This addresses the issue we are having now where releasing encounters errors:

```
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.apache.maven.plugins:maven-release-plugin:3.0.1:prepare (default-cli) on project web5-parent: Unable to commit files
Provider message:
The git-push command failed.
Command output:
remote: error: GH006: Protected branch update failed for refs/heads/main.        
remote: error: Changes must be made through a pull request.        
To https://github.com/TBD54566975/web5-kt.git
 ! [remote rejected] main -> main (protected branch hook declined)
error: failed to push some refs to 'https://github.com/TBD54566975/web5-kt.git'
```

https://github.com/TBD54566975/web5-kt/actions/runs/8593572310/job/23545262111#step:6:3973

The `tbd-releases` account should be able to push to `main` as defined by the config:

<img width="810" alt="image" src="https://github.com/TBD54566975/web5-kt/assets/199891/404cb829-8181-4b19-9ada-81464e5d0861">

Searching through audit log I discovered why this is failing:

<img width="546" alt="image" src="https://github.com/TBD54566975/web5-kt/assets/199891/3c10283d-593f-4309-bbf1-f0efe69ef9cf">

Note `actorId` of `41898282`. That maps to the `github-actions[bot]`: https://api.github.com/user/41898282

We have been thinking that we were trying to push using the `tbd-releases` Personal Access Token, and it's not.

To fix, set this token in https://github.com/actions/checkout property `token` and see how that works.

ALRubinger:
Done in 5f1424ac8e3851d715368ff03f785bdb6a8d14e5

PR #298, opened 2024-04-08
Issue #297: Check out for releases using tbd-releases PAT

ALRubinger:
Address #297


ALRubinger:
Force merging this through as it is the only way to test, and unblock teams for release ahead of start of work Monday.

PR #306, opened 2024-05-01
Add VC signing & verifying binded code

KendallWeihe:
Not intended for review, this is a part of a spike to prove out retrofitting bindings. This PR includes changes to make use of the bindings made available here https://github.com/TBD54566975/web5-rs/pull/139 and here https://github.com/TBD54566975/web5-rs/pull/140

### DIFF
--- a/credentials/src/main/kotlin/web5/sdk/credentials/VerifiableCredential.kt
+++ b/credentials/src/main/kotlin/web5/sdk/credentials/VerifiableCredential.kt
@@ -85,8 +85,6 @@ public class VerifiableCredential internal constructor(public val vcDataModel: V
     val coreBearerDid = runBlocking {
       bearerDidFromKeyManager(did.uri, did.keyManager.getCore())
     }
-    println(coreBearerDid)
-    println(did.document.verificationMethod)
     val contexts = vcDataModel.contexts.map { it.toString() }
     val id = vcDataModel.id.toString()
     val types = vcDataModel.types
@@ -95,10 +93,7 @@ public class VerifiableCredential internal constructor(public val vcDataModel: V
     val expirationDate = vcDataModel.expirationDate?.time
     val credentialSubject = web5.sdk.core.CredentialSubject(vcDataModel.credentialSubject.id.toString(), null)
     val coreVc = CoreVerifiableCredential(contexts, id, types, issuer, issuanceDate, expirationDate, credentialSubject)
-    println(coreVc)
     val signedJwt = coreVc.sign(coreBearerDid, KeySelector.MethodType(VerificationMethodType.VERIFICATION_METHOD))
-    println(signedJwt)
-
     return signedJwt
   }
 
--- a/credentials/src/main/kotlin/web5/sdk/credentials/VerifiableCredential.kt
+++ b/credentials/src/main/kotlin/web5/sdk/credentials/VerifiableCredential.kt
@@ -223,59 +223,6 @@ public class VerifiableCredential internal constructor(public val vcDataModel: V
       runBlocking {
         verifyJwt(vcJwt)
       }
-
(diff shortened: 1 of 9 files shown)


pom.xml
```diff
@@ -129,6 +129,12 @@
   <dependencyManagement>
     <!-- Direct Dependencies -->
     <dependencies>
+      <dependency>
```
KendallWeihe, line 132: figured out we could put this dependency at the root level

dids/src/main/kotlin/web5/sdk/dids/methods/jwk/DidJwk.kt
```diff
@@ -52,7 +54,7 @@ public object DidJwk {
     keyManager: KeyManager = InMemoryKeyManager(),
     algorithmId: AlgorithmId = AlgorithmId.Ed25519): BearerDid {
 
-    val keyAlias = keyManager.generatePrivateKey(algorithmId)
+    val keyAlias = keyManager.generatePrivateKey(algorithmId, DefaultKeyGenOptions("0"))
```
KendallWeihe, line 57: implement key gen options so that we can define the key alias at generation time

crypto/src/main/kotlin/web5/sdk/crypto/KeyManager.kt
```diff
@@ -14,6 +15,8 @@ import web5.sdk.crypto.jwk.Jwk
  */
 public interface KeyManager {
 
+  fun getCore(): KeyManagerCore
```
KendallWeihe, line 18: this is going to be a common practice where we expose new functions to return the "core" type for the given type

credentials/src/test/kotlin/web5/sdk/credentials/VerifiableCredentialTest.kt
```diff
@@ -217,9 +212,8 @@ fun `kyc credential verify does not throw an exception if vc is legit`() {
 
   @Test
   fun `verify does not throw an exception if vc signed with did dht is legit`() {
-    val keyManager = InMemoryKeyManager()
-    val issuerDid = DidDht.create(keyManager)
-    val holderDid = DidDht.create(keyManager)
+    val issuerDid = DidDht.create(InMemoryKeyManager())
```
KendallWeihe, line 215: any did:dht tests will currently fail

credentials/src/main/kotlin/web5/sdk/credentials/VerifiableCredential.kt
```diff
@@ -209,58 +220,9 @@ public class VerifiableCredential internal constructor(public val vcDataModel: V
      * ```
      */
     public fun verify(vcJwt: String) {
-      val decodedJwt = Jwt.decode(vcJwt)
```
KendallWeihe, line 212: this function is a big win (obv the actual verification logic needs to be audited such that we aren't losing functionality, but this is a spike so doesn't matter right now)

credentials/src/main/kotlin/web5/sdk/credentials/VerifiableCredential.kt
```diff
@@ -76,14 +82,19 @@ public class VerifiableCredential internal constructor(public val vcDataModel: V
...
-      .issueTime(vcDataModel.issuanceDate.time)
-      .subject(vcDataModel.credentialSubject.id.toString())
-      .misc("vc", vcDataModel.toMap())
-      .build()
-
-    return Jwt.sign(did, payload)
+    val coreBearerDid = runBlocking {
+      bearerDidFromKeyManager(did.uri, did.keyManager.getCore())
+    }
+    val contexts = vcDataModel.contexts.map { it.toString() }
```
KendallWeihe, line 88: all of this is verbose and could be done better but it's fine for now

Issue #310, opened 2024-05-05
[Bug] Cannot update DID documents


angiejones:
**Describe the bug**
There are several reasons why one might need to update a DID document including adding/deleting/updating service endpoints. However, the fields of the DidDocument class are final, and therefore cannot be updated. 

**To Reproduce:**

Create and publish a DID:

```kt
var myBearerDid = DidDht.create(keyManager, CreateDidDhtOptions(publish = true))
```

Try to add a service later:

```kt
val serviceToAdd = Service.Builder()
      .id("pfi")
      .type("PFI")
      .serviceEndpoint(listOf("https://example.com/"))
      .build()

// this is immutable so there are no changes that can be made here
myBearerDid.document.service.
```

**Expected behavior:**
```kt
val serviceToAdd = Service.Builder()
      .id("pfi")
      .type("PFI")
      .serviceEndpoint(listOf("https://example.com/"))
      .build()

// if this were mutable, people could add a new service then republish their doc
myBearerDid.document.service.add(serviceToAdd)
DidDht.publish(keyManager, myBearerDid.document)
```

**Supporting Material**
This is possible in web5-js:

```js
const myBearerDid = await DidDht.create({ options: { publish: true } });

myBearerDid.document.service = myBearerDid.document.service || [];

// add a service to the DID document
myBearerDid.document.service.push({
  id: 'pfi',
  type: 'PFI',
  serviceEndpoint: 'https://example.com/'
});

// republish the updated DID document
DidDht.publish({did: myBearerDid});
```

michaelneale:
@angiejones as a work around: 

```kotlin
    val keyManager = InMemoryKeyManager()
    var myBearerDid = DidDht.create(keyManager, CreateDidDhtOptions(publish = true))

    val existingBearerDid: BearerDid = myBearerDid

    val serviceToAdd = Service.Builder()
      .id("pfi")
      .type("PFI")
      .serviceEndpoint(listOf("https://example.com/"))
      .build()

    val updatedServices = existingBearerDid.document.service?.toMutableList() ?: mutableListOf()
    updatedServices.add(serviceToAdd)

    print(updatedServices)

    val updatedDidDocument = DidDocument(
      id = existingBearerDid.document.id,
      context = existingBearerDid.document.context,
      controller = existingBearerDid.document.controller,
      verificationMethod = existingBearerDid.document.verificationMethod,
      authentication = existingBearerDid.document.authentication,
[...]

PR #311, opened 2024-05-07
add update for did

michaelneale:
# Overview
 For bug: https://github.com/TBD54566975/web5-kt/issues/310

need a way to easily modify a did or document. This uses the (I believe) idiomatic kotlin way which keeps the data objects immutable.

### DIFF
--- a/dids/src/main/kotlin/web5/sdk/dids/methods/dht/DidDht.kt
+++ b/dids/src/main/kotlin/web5/sdk/dids/methods/dht/DidDht.kt
@@ -244,14 +244,21 @@ public sealed class DidDhtApi(configuration: DidDhtConfiguration) {
       updatedVerificationMethods.add(verificationMethod)
     }
 
-    val updatedDidDocument = existingDidDocument.copy(
-      service = updatedServices?.toList(),
-      controller = updatedControllers?.toList(),
+    val updatedDidDocument = DidDocument(
+      existingDidDocument.id,
+      existingDidDocument.context,
       alsoKnownAs = updatedAlsoKnownAses?.toList(),
-      verificationMethod = updatedVerificationMethods
+      controller = updatedControllers?.toList(),
+      verificationMethod = updatedVerificationMethods,
+      service = updatedServices?.toList(),
+      assertionMethod = existingDidDocument.assertionMethod,
+      authentication = existingDidDocument.authentication,
+      keyAgreement = existingDidDocument.keyAgreement,
+      capabilityDelegation = existingDidDocument.capabilityDelegation,
+      capabilityInvocation = existingDidDocument.capabilityInvocation
     )
 
-    val updatedBearerDid = bearerDid.copy(document = updatedDidDocument)
+    val updatedBearerDid = BearerDid(bearerDid.uri, bearerDid.did, bearerDid.keyManager, document = updatedDidDocument)
 
     if (options.publish) {
       publish(bearerDid.keyManager, updatedDidDocument)
--- a/dids/src/main/kotlin/web5/sdk/dids/methods/dht/DidDht.kt
+++ b/dids/src/main/kotlin/web5/sdk/dids/methods/dht/DidDht.kt
@@ -99,14 +99,17 @@ public class CreateDidDhtOptions(
 
 /**
  * Specifies options for updating a "did:dht" Decentralized Identifier (DID).
- * Any options not specified will retain the value from the DID being updated.
+ * Any options not specified will retain the value from the DID being updated, otherwise
+ * the values provided will be set in the DID document.
+ * For example if you wanted to update the services, pass in the full list of `services` you
+ * would like the updated DID Document to contain.
  *
  * @property verificationMethods A list of [Jwk]s to add to the DID Document mapped to their purposes
  * as verification methods, and an optional controller for the verification method.
- * @property services A list of [Service]s to add to the DID Document.
+ * @property services A list of [Service]s that will be set in the DID Document.
  * @property publish Whether to publish the DID Document to the DHT after creation.
- * @property controllers A list of controller DIDs to add to the DID Document.
- * @property alsoKnownAses A list of also known as identifiers to add to the DID Document.
+ * @property controllers A list of controller DIDs that will be set in the DID Document.
+ * @property alsoKnownAses A list of also known as identifiers to be set in the DID Document.
  */
 public typealias UpdateDidDhtOptions = CreateDidDhtOptions
 
@@ -219,7 +222,7 @@ public sealed class DidDhtApi(configuration: DidDhtConfiguration) {
    * Updates an existing "did:dht" DID by applying the provided changes to the DID Document.
    *
    * @param bearerDid The existing "did:dht" DID to update.
-   * @param options Optional parameters ([UpdateDidDhtOptions]) to specify additional keys, services, and optional
+   * @param options Optional parameters ([UpdateDidDhtOptions]) to specify keys, services, and optional
    * publishing during the update.
    * @return The updated [BearerDid] instance.
    */
--- a/dids/src/main/kotlin/web5/sdk/dids/methods/dht/DidDht.kt
+++ b/dids/src/main/kotlin/web5/sdk/dids/methods/dht/DidDht.kt
@@ -219,7 +219,7 @@ public sealed class DidDhtApi(configuration: DidDhtConfiguration) {
    * Updates an existing "did:dht" DID by applying the provided changes to the DID Document.
    *
    * @param bearerDid The existing "did:dht" DID to update.
-   * @param options Optional parameters ([CreateDidDhtOptions]) to specify additional keys, services, and optional
+   * @param options Optional parameters ([UpdateDidDhtOptions]) to specify additional keys, services, and optional
    * publishing during the update.
    * @return The updated [BearerDid] instance.
    */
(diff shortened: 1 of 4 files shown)


dids/src/test/kotlin/web5/sdk/dids/did/BearerDidTest.kt
```diff
@@ -8,13 +8,44 @@ import org.mockito.kotlin.spy
...
+
+    val existingBearerDid: BearerDid = myBearerDid
+
+    val serviceToAdd = Service.Builder()
+      .id("pfi")
+      .type("PFI")
+      .serviceEndpoint(listOf("https://example.com/"))
+      .build()
+
+    val newServiceList = existingBearerDid.document.service.orEmpty() + serviceToAdd
```
jiyoontbd, line 36: So I think with the current impl you'll add the existing service again with this line, because you add whatever is passed in options to the existing service list? did I read the impl wrong

jiyoontbd:
This is great thanks @michaelneale ! one thing I'm wondering is if we should just make service list mutable if that is expected usage instead of trying to preserve immutability which is typical in Kotlin.

@angiejones what other things inside a did are expected to be liable to mutation after initial creation, besides service?

michaelneale:
@jiyoontbd yeah - I wasn't sure about idioms around mutability (as a rule immutable is ideal as long as DX is good enough as it eliminates a whole class of bugs).

dids/src/test/kotlin/web5/sdk/dids/did/BearerDidTest.kt
michaelneale: hrm - I think that is my mistake, intention wasn't to only add but replace - let me add better tests and fix if needed before merging

michaelneale:
PTAL @jiyoontbd again regarding your pertinent comments

dids/src/main/kotlin/web5/sdk/dids/methods/dht/DidDht.kt
```diff
@@ -202,6 +202,47 @@ public sealed class DidDhtApi(configuration: DidDhtConfiguration) {
...
 
+  /**
+   * Updates an existing "did:dht" DID by applying the provided changes to the DID Document.
+   *
+   * @param bearerDid The existing "did:dht" DID to update.
+   * @param options Optional parameters ([CreateDidDhtOptions]) to specify additional keys, services, and optional
+   * publishing during the update.
+   * @return The updated [BearerDid] instance.
+   */
+  public fun update(bearerDid: BearerDid, options: CreateDidDhtOptions): BearerDid {
```
jiyoontbd, line 213: Thoughts on calling the option type passed in here `UpdateDidDhtOptions` ?

jiyoontbd (approved):
One smol comment

dids/src/main/kotlin/web5/sdk/dids/methods/dht/DidDht.kt
michaelneale: @jiyoontbd great idea - added a public typealias for that reason (it is the same type) but with its own docs to make semantics clear. PTAL

jiyoontbd: Cool. If we can call .Create() with CreateOptions and .Update with UpdateOptions that would be ideal

michaelneale: yep, done

jiyoontbd (approved):

dids/src/main/kotlin/web5/sdk/dids/methods/dht/DidDht.kt
```diff
@@ -202,6 +215,47 @@ public sealed class DidDhtApi(configuration: DidDhtConfiguration) {
...
+   * @param options Optional parameters ([UpdateDidDhtOptions]) to specify additional keys, services, and optional
+   * publishing during the update.
+   * @return The updated [BearerDid] instance.
+   */
+  public fun update(bearerDid: BearerDid, options: UpdateDidDhtOptions): BearerDid {
+    val existingDidDocument = bearerDid.document
+
+    val updatedServices = options.services ?: existingDidDocument.service
+    val updatedControllers = options.controllers ?: existingDidDocument.controller
+    val updatedAlsoKnownAses = options.alsoKnownAses ?: existingDidDocument.alsoKnownAs
```
KendallWeihe, line 231: I think you might have a bug here with how you're using the Elvis operator (👑 🎸) (`?:`). The `UpdateDidDhtOptions` has docs:

```kotlin
 * @property services A list of [Service]s to add to the DID Document.
// ...
 * @property controllers A list of controller DIDs to add to the DID Document.
 * @property alsoKnownAses A list of also known as identifiers to add to the DID Document.
```

As in, those values are additive, but the Elvis operator is a null coalescing operation, so imagine the `existingDidDocument.service` has service A, and then `options.services` has service B & C, then this would update the did:dht's DID Document to now have only services B & C.

KendallWeihe (approved):
one comment with respect to a bug, but I trust your judgement to merge once addressed, everything else lgtm!

dids/src/main/kotlin/web5/sdk/dids/methods/dht/DidDht.kt
michaelneale: @KendallWeihe ah - that is a bug in the docs, it isn't adding now, but replacing (ie code is right, comments are not). will fix. thanks!

jiyoontbd: Hmm, is the desired behavior to replace completely or add a new service to an existing list?

If it's the former, is this because you are prioritizing immutability?

I am not certain that going with former approach is a good dx. If I just want to add a service, shouldn't the update method do the work internally of:

1. check if service method is empty
2. if not empty, turn list mutable
3. add to list
4. make the list immutable again

?

michaelneale: yeah desired is to entirely replace it (at least as it stands currently) which preserves immutability but keeps it relatively simple (but alternative means having add/remove etc things explicitly in there for things that are lists, perhaps add/remove for each thing you may want to change and separate methods?)

angiejones:
good to merge this one yet, @michaelneale?

michaelneale:
@angiejones I think @jiyoontbd wanted to have a think over it and if we go this approach or move to a mutable approach or perhaps specific functions for everything we want to update (vs an Update/Create options class).

michaelneale:
@jiyoontbd would you like me to (seperately) try a mutable take on this change that lets add/remove services? (and other things, perhaps) to see what it feels like?

mistermoe:
hmm so thinking about this a bit:

ultimately, publishing updates to a pre-existing `did:dht's` did document requires re-publishing the net new did document to mainline dht effectively clobbering the existing one.

Given ^, i see where you were headed with your `copy` approach @michaelneale. Definitely agree that while technically correct, the api surface may have not been immediately intuitive.

If we take an `update` approach the question still remains of whether `update` takes an entirely new did document OR a series of "patches" to be applied to the pre-existing did document. applying said patches to the pre-existing did document would result in a net new did document that would then be published to mainline. 

if `update` takes an entirely new `did document` then we would likely want to add methods to `DidDocument` itself, one of which would be `copy`. modify the copy, provide the modified copy to `update` and done.


if `update` takes a series of patches, we'd probably want to lean on a pre-existing standard for this e.g. [`json patch`](https://jsonpatch.com/). 


tbh, both seem a bit painful to use.

Would be helpful to think through 1 use-case for this feature to help guide the design. here's one 

> one of our private keys has been compromised whose public key is a verification method in our did doc. we need to replace it with a new one

^ requires:
1. removing the compromised verification method from the did doc
2. adding a new verification method to the did doc

if we take the patching approach or even a subset thereof (`remove` and `add` only) we would at least be able to provide the same functionality that `create` does in that it will create keys using the provided key manager for you. 


`update` taking an entire did document would require the caller to handle key creation and removal themselves.

mistermoe:
either way, if there is a need for updating `did:dht` did documents at the moment and absence of it is blocking someone, imo, we could go ahead and merge this PR in so that there's _a_ way to do this (presuming that what's in this PR works!)

generally speaking, updating did docs is certainly something we'll want to consider broadly as the need will inevitably arise for us and others

michaelneale:
@mistermoe no not blocking, so can iterate here. I like the update + patch approach (this one is close to that but does require you to pass in full things to replace it with).

michaelneale:
here is another take: https://github.com/TBD54566975/web5-kt/pull/313 - that is adding/updating/deleting just service, for comparison: 

https://github.com/TBD54566975/web5-kt/pull/313

dids/src/main/kotlin/web5/sdk/dids/methods/dht/DidDht.kt
KendallWeihe: > yeah desired is to entirely replace it

I agree with this approach, if only for the sake of pragmatism ("crawl before walk")

angiejones:
> either way, if there is a need for updating `did:dht` did documents at the moment and absence of it is blocking someone, imo, we could go ahead and merge this PR in so that there's _a_ way to do this (presuming that what's in this PR works!)
> 
> generally speaking, updating did docs is certainly something we'll want to consider broadly as the need will inevitably arise for us and others

It already has. fortunately it was easier for the PFI to do so in JS.

michaelneale:
I think I will close this once https://github.com/TBD54566975/web5-kt/pull/313 is blessed (as it is minimalist and doesn't block off any future patterns but solves the issue).

PR #313, opened 2024-05-16
another take on simpler way of adding a service

michaelneale:
For bug https://github.com/TBD54566975/web5-kt/issues/310

an alternative take on https://github.com/TBD54566975/web5-kt/pull/311 which adds a service more directly without taking an update object.

### DIFF
--- a/dids/src/main/kotlin/web5/sdk/dids/did/BearerDid.kt
+++ b/dids/src/main/kotlin/web5/sdk/dids/did/BearerDid.kt
@@ -75,27 +75,6 @@ public class BearerDid(
     return BearerDid(uri, did, keyManager, updatedDocument)
   }
 
-  /**
-   * Deletes a service from the DID Document by its ID and returns a new `BearerDid` instance with the updated document.
-   *
-   * @param serviceId The ID of the service to delete from the DID Document.
-   * @return A new `BearerDid` instance with the updated DID Document.
-   */
-  public fun deleteService(serviceId: String): BearerDid {
-    val updatedServices = document.service?.filter { it.id != serviceId } ?: emptyList()
-    val updatedDocument = createUpdatedDocument(updatedServices)
-    return BearerDid(uri, did, keyManager, updatedDocument)
-  }
-
-  /**
-   * Clears all services from the DID Document and returns a new `BearerDid` instance with the updated document.
-   *
-   * @return A new `BearerDid` instance with the updated DID Document.
-   */
-  public fun clearServices(): BearerDid {
-    val updatedDocument = createUpdatedDocument(emptyList())
-    return BearerDid(uri, did, keyManager, updatedDocument)
-  }
 
   /**
    * Creates a new `DidDocument` instance with the updated services.
--- a/dids/src/main/kotlin/web5/sdk/dids/did/BearerDid.kt
+++ b/dids/src/main/kotlin/web5/sdk/dids/did/BearerDid.kt
@@ -7,6 +7,7 @@ import web5.sdk.crypto.KeyManager
 import web5.sdk.crypto.jwk.Jwk
 import web5.sdk.dids.didcore.DidDocument
 import web5.sdk.dids.didcore.Did
+import web5.sdk.dids.didcore.Service
 import web5.sdk.dids.didcore.VMSelector
 import web5.sdk.dids.didcore.VerificationMethod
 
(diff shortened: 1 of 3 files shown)


dids/src/main/kotlin/web5/sdk/dids/did/BearerDid.kt
```diff
@@ -61,6 +62,62 @@ public class BearerDid(
...
+  }
+
+  /**
+   * Deletes a service from the DID Document by its ID and returns a new `BearerDid` instance with the updated document.
+   *
+   * @param serviceId The ID of the service to delete from the DID Document.
+   * @return A new `BearerDid` instance with the updated DID Document.
+   */
+  public fun deleteService(serviceId: String): BearerDid {
+    val updatedServices = document.service?.filter { it.id != serviceId } ?: emptyList()
```
jiyoontbd, line 85: if i don't find the serviceId, i nuke the whole list of services?

dids/src/test/kotlin/web5/sdk/dids/did/BearerDidTest.kt
```diff
@@ -1,20 +1,59 @@
...
 import web5.sdk.dids.methods.jwk.DidJwk
 import kotlin.test.Test
 import kotlin.test.assertEquals
 import kotlin.test.assertNotNull
 
 class BearerDidTest {
 
+  @Test
+  fun `addService should add a new service to the DID Document`() {
+    val bearerDid = DidJwk.create(InMemoryKeyManager())
```
jiyoontbd, line 22: nit: could we do this using DidDht.create, since that's what #310 used to describe the bug?

jiyoontbd (commented):
i'm generally more in favor of this PR than #311 , with a couple caveats:

the original issue #310 is wanting a DX that allows a service to be **added** and not a whole new list of services to be provided to be replaced. whether under the hood a whole new did doc is created / published should be irrelevant to the client, imo. 

in this PR, i'd also like to see removal of deleteService() and clearService() as that was not in the scope of #310 , and according to https://github.com/TBD54566975/web5-kt/pull/311#issuecomment-2112742508 we have a bit more thinking to do wrt how to treat `update()` as a whole.

dids/src/main/kotlin/web5/sdk/dids/did/BearerDid.kt
jiyoontbd: nvm, i'm blind. if service is null then i get an empty list

angiejones:
i like this approach

michaelneale:
ok, let me tighten it up to just have add - as per @jiyoontbd comment as that keeps it focussed, I like it.

dids/src/test/kotlin/web5/sdk/dids/did/BearerDidTest.kt
michaelneale: done

michaelneale:
@jiyoontbd PTAL again

jiyoontbd (approved):

dids/src/main/kotlin/web5/sdk/dids/did/BearerDid.kt
```diff
@@ -61,6 +62,41 @@ public class BearerDid(
...
+   *
+   * @param service The service to add to the DID Document.
+   * @return A new `BearerDid` instance with the updated DID Document.
+   */
+  public fun addService(service: Service): BearerDid {
+    val updatedServices = document.service?.toMutableList() ?: mutableListOf()
+    updatedServices.add(service)
+    val updatedDocument = createUpdatedDocument(updatedServices)
+    return BearerDid(uri, did, keyManager, updatedDocument)
+  }
```
mistermoe, line 76: would it make more sense for `addService` to be a method on `DidDocument`?  that's what we've done in [go](https://github.com/TBD54566975/web5-go/blob/main/dids/didcore/document.go#L195-L198) and [dart](https://github.com/TBD54566975/web5-dart/blob/main/packages/web5/lib/src/dids/did_core/did_document.dart#L172-L175).

this method is helpful for building did docs in general. examples [here](https://github.com/TBD54566975/web5-dart/blob/main/packages/web5/lib/src/dids/did_dht/did_dht.dart#L75-L77) and [here](https://github.com/TBD54566975/web5-dart/blob/main/packages/web5/lib/src/dids/did_web/did_web.dart#L74-L76). same usage exists in go as well

michaelneale: that would only work if the service was mutable (like it is in other places) - otherwise currently, has to return a new did (with it added to the doc) OR you need to return a new doc and then update the doc in the did (which returns a new did) if sticking with immutable approach in kotlin (correct me if wrong).

michaelneale: ie, you would do blah.document.addService()... but it wouldn't change document in place but return a new document as vaaaaaals: 

image

michaelneale: so - we could change them to be mutable lists?

jiyoontbd:
hey @michaelneale ! 

ok, so after some discussion, i'd like to propose that we close this PR for now, and address the issue #310 once we hold some design discussions on how exactly one performs a DID `update` operation.

the reason being is that even if we have `addService()` implemented, this method would become tech debt quickly because it isn't considering how each did method handles updates. 

the core issue here is that we're only mutating our own `BearerDid` abstraction by using this method. however, adding service (or updating any part of a did) should have downstream actions that we need to consider for each did method (did:dht handles updates differently than did:jwk, for example)

i've written about this a bit more as an issue in web5-spec. let me know if you have any questions! https://github.com/TBD54566975/web5-spec/issues/157

michaelneale:
makes sense to me @jiyoontbd! - will close this (maybe keep the branch around temporarily just in case).

PR #317, opened 2024-05-16
consistent artifact version for bouncycastle dependencies

jiyoontbd:


pom.xml
```diff
@@ -241,6 +241,11 @@
         <artifactId>bcpkix-jdk15to18</artifactId>
         <version>${version.org.bouncycastle}</version>
       </dependency>
+      <dependency>
+        <groupId>org.bouncycastle</groupId>
+        <artifactId>bcprov-jdk18on</artifactId>
```
leordev, line 246: Perfect! Now it seems we have `bcprov-jdk18on, bcpkix-jdk15to18, bcprov-jdk15to18` but as per @ALRubinger suggestion, can you please also add the `bcutil-jdk15to18`? Just so we make sure everything is consistent!

jiyoontbd: done

Issue #324, opened 2024-06-26
[Bug] web did resolution does not support http ports


aparkersquare:
**Describe the bug**
Resolving a web did with a port results in an incorrect url for the `did.json`

**To Reproduce:**

Using the following code
```
import web5.sdk.dids.DidResolvers

fun main(args: Array<String>) {

  val did = "did:web:127.0.0.1:8802"
  try {
    val resolution = DidResolvers.resolve(did)
    println(resolution)
  } catch (t: Throwable) {
    println("Error resolving DID: $did")
    println("  ${t.stackTraceToString()}".replace("\n", "\n  "))
  }
}
```

This returns `{"didDocumentMetadata":{},"didResolutionMetadata":{"error":"internalError"}}`

Debugging through web5-kt I see the `docUrl` is using the port as part of the path, and also leaving off the `.well-known`.

`docUrl=https://127.0.0.1/8802/did.json`

This seems to be an issue somewhere in [DidWeb.kt](https://github.com/TBD54566975/web5-kt/blob/489ca7869534ea6253a3fec2e259acce8ee1608e/dids/src/main/kotlin/web5/sdk/dids/methods/web/DidWeb.kt#L200-L200).

aparkersquare:
Update - there is already a way to encode the port, and it's even in the unit test!

Example : `did:web:[example.com](http://example.com/)%3A3000:user:alice`

There is still an issue for local testing that `DidWeb` assumes `https`.

It would be good if it switched to `http` for local addresses, similar to `dap-dart`.
See [did_web.dart#L94-L100](https://github.com/TBD54566975/web5-dart/blob/main/lib/src/dids/did_web/did_web.dart#L94-L100)  (thanks to @mistermoe for the pointer).

PR #325, opened 2024-06-26
support web did resolution over http for local host

aparkersquare:
# Overview
Uses `http` for web did resolution for local hosts. This allows for local development.

See https://github.com/TBD54566975/web5-kt/issues/324


dids/src/main/kotlin/web5/sdk/dids/methods/web/DidWeb.kt
```diff
@@ -192,15 +192,15 @@ public sealed class DidWebApi(
   private fun decodeId(parsedDid: Did): String {
```
mistermoe, line 192: @aparkersquare thoughts on using a parsed url to build `targetUrl`? something like

```kotlin
  public fun decodeId(parsedDid: Did): String {
    val domainNameWithPath = parsedDid.id.replace(":", "/")
    val decodedDomain = URLDecoder.decode(domainNameWithPath, UTF_8)
    val url = URL("https://$decodedDomain")

    val decodedId = StringBuilder()
    if (url.host == "localhost" || url.host == "127.0.0.1") {
      decodedId.append("http://")
    } else {
      decodedId.append("https://")
    }

    decodedId.append(decodedDomain)

    if (url.path.isEmpty()) {
      decodedId.append(WELL_KNOWN_URL_PATH)
    }
    decodedId.append(DID_DOC_FILE_NAME)

    return decodedDomain.toString()
  }
```

potential benefits of doing it this way are:
* url is parsed (therefore "validated" prior to doing anything else
* we can check `url.host` explicitly

there's probably kotlin idioms that can be used in the above suggestion to make the code more kotlin-esque which you'd know better about than me!

when i gave chad (aka chatGPT) the above suggestion and asked him for something more idiomatic, he suggested:

```kotlin
public fun decodeId(parsedDid: Did): String {
    val domainNameWithPath = parsedDid.id.replace(":", "/")
    val decodedDomain = URLDecoder.decode(domainNameWithPath, UTF_8.name())
    val url = URL("https://$decodedDomain")

    return buildString {
        append(if (url.host == "localhost" || url.host == "127.0.0.1") "http://" else "https://")
        append(decodedDomain)
        if (url.path.isEmpty()) {
            append(WELL_KNOWN_URL_PATH)
        }
        append(DID_DOC_FILE_NAME)
    }
}
```

let me know what you think! i personally prefer the former but maybe that's because i don't write much kotlin. Either way, 👍🏽

mistermoe (approved):
thanks @aparkersquare ! left a comment and approved regardless so you're not stuck waiting for approval

dids/src/main/kotlin/web5/sdk/dids/methods/web/DidWeb.kt
aparkersquare: Oooh, I didn't know about `url.host`!
Agree, that looks like it could be a nicer approach. Thanks.

I'll have a closer look soon.

jiyoontbd (approved):

dids/src/main/kotlin/web5/sdk/dids/methods/web/DidWeb.kt
aparkersquare: This works.
Updated the PR to use this approach, and deleted the LocalHostDomainMatcher

mistermoe: Naisuu!

angiejones:
thanks so much @aparkersquare 🙏🏾

taniashiba:
thank you so much!

Issue #322, opened 2024-06-25
Move `AwsKeyManager` into it's own module


mistermoe:
Currently, [`AwsKeyManager`](https://github.com/TBD54566975/web5-kt/blob/main/crypto/src/main/kotlin/web5/sdk/crypto/AwsKeyManager.kt) lives in `web5-crypto` module which happens to be the foundation of all other web5 modules. 

`AwsKeyManager` relies on `aws-java-sdk-kms` as a dependency which results in `web5-crypto`'s dependency tree to look like this:

```
[INFO]
[INFO] --- dependency:3.6.1:tree (default-cli) @ web5-crypto ---
[INFO] xyz.block:web5-crypto:jar:0.0.0-main-SNAPSHOT
[INFO] +- xyz.block:web5-common:jar:0.0.0-main-SNAPSHOT:compile
[INFO] +- xyz.block:web5-testing:jar:0.0.0-main-SNAPSHOT:test
[INFO] +- com.nimbusds:nimbus-jose-jwt:jar:9.37.2:compile
[INFO] |  \- com.github.stephenc.jcip:jcip-annotations:jar:1.0-1:compile
[INFO] +- com.amazonaws:aws-java-sdk-kms:jar:1.12.668:compile
[INFO] |  +- com.amazonaws:aws-java-sdk-core:jar:1.12.668:compile
[INFO] |  |  +- commons-logging:commons-logging:jar:1.1.3:compile
[INFO] |  |  +- commons-codec:commons-codec:jar:1.15:compile
[INFO] |  |  +- org.apache.httpcomponents:httpclient:jar:4.5.13:compile
[INFO] |  |  |  \- org.apache.httpcomponents:httpcore:jar:4.4.13:compile
[INFO] |  |  +- com.fasterxml.jackson.dataformat:jackson-dataformat-cbor:jar:2.12.6:compile
[INFO] |  |  \- joda-time:joda-time:jar:2.8.1:compile
[INFO] |  \- com.amazonaws:jmespath-java:jar:1.12.668:compile
[INFO] +- com.fasterxml.jackson.module:jackson-module-kotlin:jar:2.17.0:compile
[INFO] |  +- com.fasterxml.jackson.core:jackson-databind:jar:2.17.0:compile
[INFO] |  |  \- com.fasterxml.jackson.core:jackson-core:jar:2.17.0:compile
[INFO] |  +- com.fasterxml.jackson.core:jackson-annotations:jar:2.17.0:compile
[...]

PR #327, opened 2024-06-27
Issue #322: Move AwsKeyManager to its own module

ALRubinger:
## Notes to Reviewers

* Note the ⚠️  below and impact on versioning of `web5-kt` this carries via an API change.
* Please do not squash merge this PR; rebase merge is fine. We should preserve 1eb180c258d790822e77742b59d3f209dc7d6b3f so that it may be cleanly reverted as part of #328 

## In this PR

Reviewers, please assess impact of the following changes introduced by this PR:

* Create new module with `artifactId`: `web5-keymanager-aws`
* Add this new module to the `distribution`
* Move `AwsKeyManager` into it, along with `AwsKeyManagerTest`
*  ⚠️ Refactor package name of `AwsKeyManager` from `web5.sdk.crypto` to `web5.sdk.keymanager.aws`. This represents an API change and in strict terms would trigger a major version bump to a `2.x.y` series of `web5-kt` on the next release.
* Remove the `crypto` module dependency on `com.amazonaws:aws-java-sdk-kms`, the motivation behind this change and this PR.
* Add explicit `dependencyManagement` for `commons-codec:commons-codec`, needed in test scope by the `crypto` module which is no longer getting this transitively via `com.amazonaws:aws-java-sdk-kms`
* Add explicit `dependencyManagament` for `org.apache.httpcomponents:httpcore` and explicit declaration on it in `dids` module, which is no longer receiving this transitively since removing the `com.amazonaws:aws-java-sdk-kms` via `crypto`.

### DIFF
--- a/credentials/pom.xml
+++ b/credentials/pom.xml
@@ -108,6 +108,12 @@
       if desired.
     -->
     <dependency>
+      <groupId>xyz.block</groupId>
+      <artifactId>web5-keymanager-aws</artifactId>
+      <version>${project.version}</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
       <groupId>io.ktor</groupId>
       <artifactId>ktor-client-mock-jvm</artifactId>
     </dependency>
--- a/credentials/src/test/kotlin/web5/sdk/credentials/VerifiableCredentialTest.kt
+++ b/credentials/src/test/kotlin/web5/sdk/credentials/VerifiableCredentialTest.kt
@@ -9,7 +9,6 @@ import org.junit.jupiter.api.assertDoesNotThrow
 import web5.sdk.common.Convert
 import web5.sdk.common.Json
 import web5.sdk.crypto.AlgorithmId
-import web5.sdk.crypto.AwsKeyManager
 import web5.sdk.crypto.InMemoryKeyManager
 import web5.sdk.crypto.Jwa
 import web5.sdk.dids.did.BearerDid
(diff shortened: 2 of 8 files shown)


mistermoe:
@ALRubinger lmk what you think about this rationale:

i think we should go with:

> Remove the kover-maven-plugin enforcement on this module alone so that the build passes even without coverage (recommend against this)

reason being,
* we're no worse off than we were before
* we benefit from a dependency tree with less depth sooner
* we should definitely create an issue to set up proper tests for `AwsKeyManager` with a test harness.  We can set up [localstack](https://www.localstack.cloud/) for this, but it will definitely take non-negligible time to do.

ALRubinger:
@mistermoe OKee, short term salve, new issue incoming for medium term proper fix (tests).

michaelneale:
I like this. And I don't think Aws key manager was used in anger just yet? so risk of api change for it is low. With the distribution, does this make pulling in aws stuff optional now?

ALRubinger:
> I like this. And I don't think Aws key manager was used in anger just yet? so risk of api change for it is low. With the distribution, does this make pulling in aws stuff optional now?

1. So I did add this to the distribution `pom`, so it's included by default when bringing in everything. If we want to treat this as an extension, it doesn't need to be in the distro. What this enables so far is for folks to bring in things piecemeal ie. `dids`, `crypto`, etc. 

2. Technically it is a published API change which would trigger a major version update, but we could make statements about `web5-kt` thus far that treat the entire `1.x.y` series as open to breaking API changes like this while the project reaches API stability over time. Project leads can make that call, but yeah - as a dev if upgraded within a major version and it broke the API I'd be confused as to why.

mistermoe:
I think we should exclude it from the default distro @ALRubinger. This should, I think, make web5-kt default rollup distro usable on android? You and @michaelneale would know much better than I

michaelneale:
I would be ok with removing it I think as part of distro - if that is ok? and we can document how to bring it in if you need it.

ALRubinger:
Great; I've got no opinions on it - really is about what the expectations are for the distro. An Android distro would exclude it anyway and add in new Android things.

I just pushed a commit to exclude it for now and we'll have to document alongside DevRel how to add this in when wanting `AwsKeyManager` support as an extension of sorts. cc: @blackgirlbytes @acekyd @angiejones

michaelneale:
yeah I think when we are ready to have flavoured distros like android/cloud can look at it then? but for now I think clear docs and examples are great. Anyone doing AWS dev AND using java would have no issues copy/paste into their pom/gradle I think

mistermoe (approved):

jiyoontbd (approved):

michaelneale (approved):

tomdaffurn:
Yep, super happy to have AWS as an optional extension. It's far from the common usage, and adds a lot of weight

tomdaffurn (approved):

PR #334, opened 2024-08-13
Fix DID DHT NPE

### LINKED ISSUE #333: [Bug] NPE thrown when no verification method id in the DNS packet
The SDK is unable to resolve a valid DID that does not contain a verification method ID in the DNS packet.

**To Reproduce:**

```kt
    val pfiDid = "did:dht:ozn5c51ruo7z63u1h748ug7rw5p1mq3853ytrd5gatu9a8mm8f1o"
    val resolution = DidDht.resolve(pfiDid)
    println("resolution: $resolution")
```

returns:

`resolution: {"didDocumentMetadata":{},"didResolutionMetadata":{"error":"internalError"}}`

**Expected behavior:**

Successful resolution. Resolving this same DID in tbdex-js works.

from @decentralgabe:
> the code does not handle the case where there’s no verification method id in the DNS packet
this is a valid response, as the id can be omitted for the identity key (verification method encoded in the DID identifier itself)

**Supporting Material**

failing [here](https://github.com/TBD54566975/web5-kt/blob/df3a1d1781e99b16d1587bea6e9b5fe37374c46a/dids/src/main/kotlin/web5/sdk/dids/methods/dht/DidDht.kt#L282) with NPE. in the screenshot, `data["id"]` is null in `DidDht:handleVerificationMethods()`

```kt
    fromDnsPacket(did, dnsPacket).let { (didDocument, types) ->
      return DidResolutionResult(
        didDocument = didDocument,
        didDocumentMetadata = DidDhtDocumentMetadata(types = types.map { it.index })
      )
    }
```

image

**Additional context**
Hackathon users are running into this. If possible, can we get a hotfix on SDK version 2.0 that doesn't include the upgrade to protocol 2.0?

decentralgabe:
Fix #333


dids/src/test/kotlin/web5/sdk/dids/methods/dht/DidDhtTest.kt
```diff
@@ -164,6 +164,20 @@ class DidDhtTest {
       }
     }
 
+    @Test
```
decentralgabe, line 167: I included this to prove it passes. I'll remove it and push again after CI succeeds

decentralgabe: update: I captured the known hex response so this test can work without relying on the DID being published

michaelneale:
nice catch 
image

dids/src/main/kotlin/web5/sdk/dids/methods/dht/DidDht.kt
```diff
@@ -639,6 +638,12 @@ public sealed class DidDhtApi(configuration: DidDhtConfiguration) {
       else -> throw IllegalArgumentException("Unknown key type: ${data["t"]}")
     }
 
+    val verificationMethodId = when {
+      data.containsKey("id") -> data["id"]!!
+      isIdentityKey(did, keyBytes) -> "0"
+      else -> publicKeyJwk.computeThumbprint()
```
jiyoontbd, line 644: do we have test cases to cover all 3 conditions in here?

decentralgabe: probably not. if we implement the [test vectors from the spec](https://did-dht.com/#test-vectors) we will be fine...

decentralgabe: i think this is a sufficient short term fix, we can take test vectors in a follow up

dids/src/main/kotlin/web5/sdk/dids/methods/dht/DhtClient.kt
```diff
@@ -81,6 +81,7 @@ internal class DhtClient(
    * @throws PkarrRecordNotFoundException if the record is not found.
    * @throws PkarrRecordResponseException if the response from the dht gateway is not successful.
    */
+  @OptIn(ExperimentalStdlibApi::class)
```
mistermoe, line 84: what's this needed for?

decentralgabe: had it in here to capture the hex response from the problem DID
can remove now

dids/src/main/kotlin/web5/sdk/dids/methods/dht/DidDht.kt
jiyoontbd: okey dokey wrote a new issue for this https://github.com/TBD54566975/web5-kt/issues/335

jiyoontbd (approved):

dids/src/main/kotlin/web5/sdk/dids/methods/dht/DidDht.kt
decentralgabe: it's still good to have that issue to make sure we implement the test vectors, but since codecov was yelling at me I added a new test case so these are all covered https://github.com/TBD54566975/web5-kt/pull/334/commits/2e29bae3bcda7336db0d7790766d07cb6c16451e

PR #337, opened 2024-08-13
dnsjava version bump

jiyoontbd:
# Overview
bump dnsjava to 3.6.0


KendallWeihe (approved):
making sure it was intentional @jiyoontbd you bumped the web5-spec submodule too fyi, I don't see any problem with it, but don't see it in the description so making sure you didn't do it by accident or if so that you're okay with it

nitro-neal (approved):

jiyoontbd:
> making sure it was intentional @jiyoontbd you bumped the web5-spec submodule too fyi, I don't see any problem with it, but don't see it in the description so making sure you didn't do it by accident or if so that you're okay with it

Yep. I bumped it to latest I could given that anything after that commit fails test due to a new select credential test vector

